PR #5, opened 2024-10-31
Add support for polymorphic serialization

AArnott:
This adds `UnionAttribute`, which can be applied to a type declaration to allow derivative types to appear at runtime where the base type is the declared type. In such a case, the data provided by the attribute dictates how the derived type will be identified in the serialized stream.

### DIFF
--- a/src/Nerdbank.MessagePack/StandardVisitor.cs
+++ b/src/Nerdbank.MessagePack/StandardVisitor.cs
@@ -1,15 +1,17 @@
 ﻿// Copyright (c) Andrew Arnott. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Frozen;
+using System.Runtime.CompilerServices;
 using System.Text;
 
 namespace Nerdbank.MessagePack;
 
 /// <summary>
-/// A <see cref="TypeShapeVisitor"/> that produces <see cref="IMessagePackConverter{T}"/> instances for each type shape it visits.
+/// A <see cref="TypeShapeVisitor"/> that produces <see cref="MessagePackConverter{T}"/> instances for each type shape it visits.
 /// </summary>
 /// <param name="owner">The serializer that created this instance. Usable for obtaining settings that may influence the generated converter.</param>
-internal class StandardVisitor(MessagePackSerializer owner) : TypeShapeVisitor
+internal class StandardVisitor(MessagePackSerializer owner) : TypeShapeVisitor, ITypeShapeFunc
 {
 	private readonly TypeDictionary converters = new();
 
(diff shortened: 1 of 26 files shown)


src/Nerdbank.MessagePack/StandardVisitor.cs
```diff
@@ -246,14 +256,75 @@ internal class StandardVisitor(MessagePackSerializer owner) : TypeShapeVisitor
...
+		if (unionAttributes is null or { Length: 0 })
+		{
+			return null;
+		}
+
+		Dictionary<int, IMessagePackConverter> deserializerData = new();
+		Dictionary<Type, (int Alias, IMessagePackConverter Converter)> serializerData = new();
+		foreach (KnownSubTypeAttribute unionAttribute in unionAttributes)
+		{
+			IObjectTypeShape? subtypeShape = (IObjectTypeShape?)objectShape.Provider.GetShape(unionAttribute.SubType);
```
eiriktsarpalis, line 301: I expect this cast will fail if the subtype happens to implement `IEnumerable`.

test/Nerdbank.MessagePack.Tests/KnownSubTypeTests.cs
```diff
@@ -0,0 +1,67 @@
...
+	public void DerivedA_AsBaseType() => this.AssertRoundtrip<BaseClass>(new DerivedA { BaseClassProperty = 5, DerivedAProperty = 6 });
+
+	[Fact]
+	public void DerivedAA_AsBaseType() => this.AssertRoundtrip<BaseClass>(new DerivedAA { BaseClassProperty = 5, DerivedAProperty = 6 });
+
+	[Fact]
+	public void DerivedB_AsBaseType() => this.AssertRoundtrip<BaseClass>(new DerivedB(10) { BaseClassProperty = 5 });
+
+	[GenerateShape]
+	[KnownSubType(1, typeof(DerivedA))]
```
eiriktsarpalis, line 44: Perhaps you could introduce a degree of type safety here by making the attribute generic:
```csharp
public class KnownSubTypeAttribute<TSubType>(int tag) where TSubType : IShapeable<TSubType>
```

eiriktsarpalis: The downside is that it forces users to apply the source generator.

test/Nerdbank.MessagePack.Tests/KnownSubTypeTests.cs
```diff
@@ -0,0 +1,67 @@
...
+	[KnownSubType(2, typeof(DerivedAA))]
+	[KnownSubType(3, typeof(DerivedB))]
+	public partial record BaseClass
+	{
+		public int BaseClassProperty { get; set; }
+	}
+
+	[GenerateShape]
```
eiriktsarpalis, line 52: It might make sense to incorporate test coverage for the case where either the base class or derived class are generic.

src/Nerdbank.MessagePack/StandardVisitor.cs
AArnott: You are right about that. I've added a test for this and will get this fixed.

test/Nerdbank.MessagePack.Tests/KnownSubTypeTests.cs
AArnott: I'm planning on writing analyzers to apply constraints to this type, including ensuring that it appears only once and derives from the type the attribute is applied to.
I'm divided on whether to require the source generator. I've recently added Serialize method overloads that allow providing a shape as an ordinary argument, so I'm kinda moving in the direction of supporting runtime supplied shapes. But not sure yet... we'll see.

PR #8, opened 2024-11-01
Add support for async (de)serialization

AArnott:
The mechanics of how it works is protected by `[Experimental]` attributes since it may be up for massive changes as we learn how to do this in a performant and usable way for Converter authors.
 
Perf comparisons have some really disappointing surprises but also some encouraging numbers.

## Arrays

| Method            | Mean      | Error    | StdDev   |
|------------------ |----------:|---------:|---------:|
| Sync_Serialize    |  72.25 μs | 0.146 μs | 0.129 μs |
| Sync_Deserialize  |  43.20 μs | 0.201 μs | 0.178 μs |
| Async_Serialize   | 209.77 μs | 0.597 μs | 0.559 μs |
| Async_Deserialize |  45.01 μs | 0.223 μs | 0.208 μs |

## Poco

| Method            | Mean       | Error   | StdDev  |
|------------------ |-----------:|--------:|--------:|
| Sync_Serialize    |   168.6 ns | 0.79 ns | 0.74 ns |
| Sync_Deserialize  |   172.1 ns | 0.38 ns | 0.36 ns |
| Async_Serialize   |   283.2 ns | 1.26 ns | 1.12 ns |
| Async_Deserialize | 1,468.1 ns | 4.40 ns | 3.90 ns |

### DIFF
--- a/src/Nerdbank.MessagePack/Converters/ArrayConverter`1.cs
+++ b/src/Nerdbank.MessagePack/Converters/ArrayConverter`1.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Andrew Arnott. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Diagnostics.CodeAnalysis;
+
 namespace Nerdbank.MessagePack.Converters;
 
 /// <summary>
@@ -9,6 +11,9 @@ namespace Nerdbank.MessagePack.Converters;
 /// <typeparam name="TElement">The element type.</typeparam>
 internal class ArrayConverter<TElement>(MessagePackConverter<TElement> elementConverter) : MessagePackConverter<TElement[]>
 {
+	/// <inheritdoc/>
+	public override bool PreferAsyncSerialization => true;
+
 	/// <inheritdoc/>
 	public override TElement[]? Deserialize(ref MessagePackReader reader, SerializationContext context)
 	{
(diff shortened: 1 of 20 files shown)


src/Nerdbank.MessagePack/Converters/ArrayConverter`1.cs
```diff
@@ -44,4 +50,75 @@ public override void Serialize(ref MessagePackWriter writer, ref TElement[]? val
...
+		cancellationToken.ThrowIfCancellationRequested();
+
+		if (value is null)
+		{
+			WriteNil(pipeWriter);
+			return;
+		}
+
+		context.DepthStep();
+		if (elementConverter.PreferAsyncSerialization)
```
AArnott, line 66: Instead of a runtime switch that changes most of the method, this could be a derived type that implements `SerializeAsync` differently.

src/Nerdbank.MessagePack/Converters/ObjectMapConverter`1.cs
```diff
@@ -23,11 +29,54 @@ public override void Serialize(ref MessagePackWriter writer, ref T? value, Seria
...
+
+		context.DepthStep();
+		SerializeHeader();
+		if (serializable.Properties is not null)
+		{
+			foreach (var property in serializable.Properties)
+			{
+				SerializePropertyName(property.RawPropertyNameString);
+				await property.WriteAsync(value, pipeWriter, context, cancellationToken).ConfigureAwait(false);
+				await FlushIfAppropriateAsync(pipeWriter, context, cancellationToken).ConfigureAwait(false);
```
AArnott: This is requiring even primitive writes (e.g. `int`) to take a `PipeWriter` directly and all the overhead that entails.
Some ideas to mitigate:

1. We could recognize 'clusters' of non-async-preferred converters and allocate a `MessagePackWriter` for each cluster and call their sync APIs instead.
1. An object that does not have an async converter among its property types should not itself have an async converter. Array converters will always prefer async since that can scale to large data sets. And any object converter with an array property will therefore be async. Then walking up the tree, any object that contains an object that contains an array will also be async. Thus, subsections of the object graph that do not 'scale' will be invoked synchronously.

src/Nerdbank.MessagePack/Converters/ArrayConverter`1.cs
eiriktsarpalis, line 66: I suspect it might be possible to avoid this virtual call if you can instead check it at folding time.

AArnott (commented):
```

BenchmarkDotNet v0.14.0, Windows 11 (10.0.26100.2033)
Unknown processor
.NET SDK 8.0.403
  [Host]     : .NET 8.0.10 (8.0.1024.46610), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.10 (8.0.1024.46610), X64 RyuJIT AVX2


```
| Method            | Mean      | Error    | StdDev    | Median    |
|------------------ |----------:|---------:|----------:|----------:|
| Sync_Serialize    |  54.49 μs | 1.134 μs |  3.343 μs |  54.78 μs |
| Sync_Deserialize  |  30.71 μs | 1.216 μs |  3.527 μs |  29.34 μs |
| Async_Serialize   | 393.15 μs | 7.771 μs | 15.873 μs | 392.52 μs |
| Async_Deserialize |  80.56 μs | 0.497 μs |  0.441 μs |  80.59 μs |

src/Nerdbank.MessagePack/Converters/ArrayConverter`1.cs
AArnott: Yes, I left a comment to that effect on this same line... :)

Unless I'm misunderstanding your suggestion...?

PR #11, opened 2024-11-02
Fix `InvalidCastException` when union subtypes implement `IEnumerable`

AArnott:
Fixes [the bug @eiriktsarpalis predicted](https://github.com/AArnott/Nerdbank.MessagePack/pull/5#discussion_r1825721992).

### DIFF
--- a/test/Nerdbank.MessagePack.Tests/KnownSubTypeTests.cs
+++ b/test/Nerdbank.MessagePack.Tests/KnownSubTypeTests.cs
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Andrew Arnott. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-public partial class UnionTests(ITestOutputHelper logger) : MessagePackSerializerTestBase(logger)
+public partial class KnownSubTypeTests(ITestOutputHelper logger) : MessagePackSerializerTestBase(logger)
 {
 	[Fact]
 	public void BaseType()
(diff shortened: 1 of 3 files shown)


test/Nerdbank.MessagePack.Tests/KnownSubTypeTests.cs
```diff
@@ -64,4 +73,12 @@ public partial record DerivedAA : DerivedA
 	public partial record DerivedB(int DerivedBProperty) : BaseClass
 	{
 	}
+
+	[GenerateShape]
+	public partial record EnumerableDerived(int Count) : BaseClass, IEnumerable<int>
```
AArnott, line 78: @eiriktsarpalis I'm not sure how often the scenario of derived subtypes that implement `IEnumerable<T>` will come up in practice, but the effect of this included some unintuitive outcomes. The properties on the type were dropped -- it serialized only as an array. And it couldn't deserialize because the constructor doesn't take an enumerable.

What happens for tree types? e.g. consider a data structure that resembles a tree. Each nodes has both children and properties. If the children are only in a collection as a property on the node, things are fine. But if the node itself implements `IEnumerable<Node>`, it seems this will malfunction. Maybe most/all serialization libraries will malfunction in that case, and it's just a bad idea for any type besides a pure collection to implement that interface. Thoughts?

eiriktsarpalis: >  I'm not sure how often the scenario of derived subtypes that implement IEnumerable<T> will come up in practice

It might be reasonable to decide that polymorphism is only supported for object shapes, and fail if any of the derived types happen to implement IEnumerable. System.Text.Json does support heterogeneous derived types (it does so by nesting the array in an envelope object containing metadata) but arguably that's overkill.

> and it's just a bad idea for any type besides a pure collection to implement that interface.

How can we distinguish between a pure collection and one that just happens to implement IEnumerable?

AArnott: I don't know. Maybe because the type has mutable properties? Pure collections may be readonly properties (e.g. Count), but won't (often?) offer property setters. I guess if a type is both enumerable and has mutable properties, it carries more data than just the enumerated state.

eiriktsarpalis: List<T> does have a settable property, Capacity. I'm sure there will be others especially given that you're free to inherit from most of the common collection types.

I'm thinking it might make sense to let users manually pick a shape kind for a type using an attribute annotation when the default convention isn't appropriate.

PR #13, opened 2024-11-02
 Add support for closed generic types in `KnownSubTypeAttribute`

AArnott:
This in a breaking change for existing users, as the sub-type is now a generic type argument to the attribute instead of a constructor argument.


test/Nerdbank.MessagePack.Tests/KnownSubTypeTests.cs
```diff
@@ -50,11 +50,16 @@ public void EnumerableDerived_BaseType()
...
 
+	[Fact]
+	public void ClosedGenericDerived_BaseType() => this.AssertRoundtrip<BaseClass>(new DerivedGeneric<int>(5) { BaseClassProperty = 10 });
+
 	[GenerateShape]
 	[KnownSubType(1, typeof(DerivedA))]
 	[KnownSubType(2, typeof(DerivedAA))]
 	[KnownSubType(3, typeof(DerivedB))]
 	[KnownSubType(4, typeof(EnumerableDerived))]
+	[KnownSubType(5, typeof(DerivedGeneric<int>))]
```
AArnott: @eiriktsarpalis this results in tests failing with

> System.InvalidOperationException : The type KnownSubTypeTests+BaseClass has a union attribute that references a type that is not known to the serializer: KnownSubTypeTests+DerivedGeneric`1[[System.Int32, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].

Which is odd, since I even deliberately added `[GenerateShape<DerivedGeneric<int>>]` below. I've tried a few variations of this but I can never get a shape of a generic type generated. Suggestions?

eiriktsarpalis: Ah yes, a witness type declaring shapes for other types generally defines its own closed typeshape provider, whereas `GenerateShape` declaration uses an implicit provider that is scoped to the entire assembly.

This is largely a historical accident, initial versions of the source generator only worked with witness type approach (largely copying the model of STJ) and `GenerateShape` was implemented on top of that infrastructure.

It might be possible that we could consolidate both approaches to use a global provider, athough I haven't thought much about the consequences of such a change.

AArnott: @eiriktsarpalis  What would you suggest so I can get a compiling test of generic subtypes?

eiriktsarpalis: You could try putting the entire hierarchy in a single witness and consume that.

PR #26, opened 2024-11-05
Promote analyzer doc to released section

AArnott:


AArnott:
The roslyn analyzer doesn't like prereleases used in the shipped file: https://github.com/dotnet/roslyn-analyzers/issues/5481#issuecomment-2458046609

PR #42, opened 2024-11-06
Use TypeShape.Roslyn in analyzer

AArnott:
The TypeShape.Roslyn.dll provides some analysis that we should avoid repeating and trying to keep in sync with TypeShape policies.

### DIFF
--- a/src/Nerdbank.MessagePack.Analyzers/KeyAttributeUseAnalyzer.cs
+++ b/src/Nerdbank.MessagePack.Analyzers/KeyAttributeUseAnalyzer.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Andrew Arnott. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using TypeShape.Roslyn;
+
 namespace Nerdbank.MessagePack.Analyzers;
 
 [DiagnosticAnalyzer(LanguageNames.CSharp)]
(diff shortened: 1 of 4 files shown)


AArnott:
My tests are now all failing with:

```
System.IO.FileNotFoundException: Could not load file or assembly 'TypeShape.Roslyn, Version=0.14.1.5139, Culture=neutral, PublicKeyToken=4846fa8de4645439'. The system cannot find the file specified.
File name: 'TypeShape.Roslyn, Version=0.14.1.5139, Culture=neutral, PublicKeyToken=4846fa8de4645439'
   at Nerdbank.MessagePack.Analyzers.KeyAttributeUseAnalyzer.<>c__DisplayClass8_0.<Initialize>b__0(CompilationStartAnalysisContext context)
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.<>c.<ExecuteCompilationStartActions>b__41_0(ValueTuple`2 data)
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.ExecuteAndCatchIfThrows_NoLock[TArg](DiagnosticAnalyzer analyzer, Action`1 analyze, TArg argument, Nullable`1 info, CancellationToken cancellationToken)
```

@sharwell do you know how I can add this external dll (comes from a nuget package) as an analyzer to the compilation to resolve this?

src/Nerdbank.MessagePack.Analyzers/KeyAttributeUseAnalyzer.cs
```diff
@@ -50,6 +52,13 @@ public override void Initialize(AnalysisContext context)
 
 		context.RegisterCompilationStartAction(context =>
 		{
+			TypeShape.Roslyn.KnownSymbols typeShapeSymbols = new(context.Compilation);
+			TypeDataModelGenerator typeDataModelGenerator = new(context.Compilation.Assembly, typeShapeSymbols, context.CancellationToken);
```
AArnott, line 56: Is this the right place to do this?

eiriktsarpalis: Seems about right. One thing to note about `TypeDataModelGenerator` is that it has no knowledge of the `GenerateShape` entrypoints or any of the other customization attributes that typeshape proper uses. These have been added in the source generator by extending that class here:

https://github.com/eiriktsarpalis/typeshape-csharp/blob/0b03b2ac2968e39885ba6dfa217b4672bc357cc1/src/TypeShape.SourceGenerator/Parser/Parser.cs

PR #46, opened 2024-11-06
Provide custom converters a means to delegate sub-values

### LINKED ISSUE #32: Custom converters should have access to other converters
A custom converter defined by a user must have the ability to delegate serialization of sub-values to other converters. This should be possible without calling out to top-level `MessagePackSerializer` methods, since that would at least reset the stack guard and possibly not reuse other options the user has set on the `MessagePackSerializer`.

AArnott:
Closes #32

### DIFF
--- a/src/Nerdbank.MessagePack/MessagePackSerializer.cs
+++ b/src/Nerdbank.MessagePack/MessagePackSerializer.cs
@@ -163,7 +163,7 @@ public record MessagePackSerializer
 	/// <param name="writer">The msgpack writer to use.</param>
 	/// <param name="value">The value to serialize.</param>
 	/// <param name="shape">The shape of <typeparamref name="T"/>.</param>
-	public void Serialize<T>(ref MessagePackWriter writer, T? value, ITypeShape<T> shape) => this.GetOrAddConverter(shape).Serialize(ref writer, ref value, this.StartingContext);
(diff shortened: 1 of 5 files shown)


src/Nerdbank.MessagePack/MessagePackSerializer.cs
```diff
@@ -77,6 +78,11 @@ public record MessagePackSerializer
 	/// </summary>
 	public SerializationContext StartingContext { get; init; } = new();
 
+	/// <summary>
+	/// Gets the shape providers that may be called upon for types that must be serialized at runtime.
+	/// </summary>
+	public ImmutableArray<ITypeShapeProvider> ShapeProviders { get; init; } = [];
```
AArnott: @eiriktsarpalis  What can I provide here as a default so that `[GenerateShape]`-attributed types will be found?
I suppose folks should have to opt in for reflection-based. But I'd like to enable types with that attribute to be automatically discovered if possible. Do I have to expose methods for those cases with generic type constraints to ensure that they implement `IShapeable<T>` in order for this to work?

AArnott: I went with another strong-typed approach, which requires that the converter itself provide the type shape (via an `IShapeable<T>` type constraint).

AArnott: But I *am* curious whether there's a use case for a `ImmutableArray<ITypeShapeProvider> ShapeProviders` member on the serializer. Maybe for the .NET Framework build where `IShapeable<T>` doesn't exist?

PR #54, opened 2024-11-08
Add `NBMsgPack031` analyzer for =1 structure being read/written in a custom converter

AArnott:
A custom converter should always write or read exactly one msgpack structure. When multiple values need to be read or written, an array or map header must be used first to introduce the number of elements that belong to one super-structure.

Conditional branching, loops and `TryRead` methods can complicate static analysis in counting how many structures are actually read/written. I'm using roslyn's control flow analysis to help out. This API doesn't have any real samples from what I can tell, so I figured this out myself and I probably didn't get it quite right yet. I have several tests that show that at least certain cases work.

There are complex cases that the analyzer doesn't handle, which the design is to just bail out without reporting a diagnostic. So this analyzer does not guarantee correct code. But it will catch most novice mistakes.

### DIFF
--- a/src/Nerdbank.MessagePack.Analyzers/ConverterAnalyzers.cs
+++ b/src/Nerdbank.MessagePack.Analyzers/ConverterAnalyzers.cs
@@ -37,7 +37,11 @@ public class ConverterAnalyzers : DiagnosticAnalyzer
 
 	public override void Initialize(AnalysisContext context)
 	{
-		context.EnableConcurrentExecution();
+		if (!Debugger.IsAttached)
+		{
+			context.EnableConcurrentExecution();
+		}
+
 		context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
 
 		context.RegisterCompilationStartAction(context =>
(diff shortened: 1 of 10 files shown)


AArnott:
@sharwell I would *love* to get your review of this, or that from another on the roslyn team familiar with control flow analysis use.

src/Nerdbank.MessagePack.Analyzers/ConverterAnalyzers.cs
```diff
@@ -67,10 +80,135 @@ public override void Initialize(AnalysisContext context)
...
+	{
+		if (context.OwningSymbol is not IMethodSymbol { IsOverride: true } method)
+		{
+			return;
+		}
+
+		int? GetVariableImpact(IInvocationOperation invocation) => invocation.Arguments.Length > 0 && invocation.Arguments[0].Value is ILiteralOperation { ConstantValue.Value: int count } ? -count : null;
+
+		INamedTypeSymbol applicableStruct;
+		Func<IInvocationOperation, (int? Impact, bool Unconditional)> relevantMethodTest;
```
AArnott, line 105: This test also needs to match on any call to another `MessagePackConverter<T>` serialize/deserialize method.

Issue #33, opened 2024-11-06
Security mitigations for hash collision attacks


AArnott:
[Our current mitigation](https://aarnott.github.io/Nerdbank.MessagePack/docs/security.html#hash-collisions) for hash collision attacks makes it possible, but not automatic. We should work with @eiriktsarpalis on https://github.com/eiriktsarpalis/typeshape-csharp/issues/33 to get a better story in place.

PR #57, opened 2024-11-09
Offer automatic secure hash functions

AArnott:
This makes it easy to use secure, hash-collision resistant equality comparer with arbitrary types for safely deserializing data.

This improves the #33 story, but doesn't close it.

### DIFF
--- a/src/Nerdbank.MessagePack/ByValueEqualityComparer`2.cs
+++ b/src/Nerdbank.MessagePack/ByValueEqualityComparer`2.cs
@@ -1,0 +1,85 @@
+﻿// Copyright (c) Andrew Arnott. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Nerdbank.MessagePack.SecureHash;
+
+namespace Nerdbank.MessagePack;
+
+/// <summary>
+/// Provides deep by-value implementations of <see cref="IEqualityComparer{T}"/> for arbitrary data types.
+/// </summary>
+/// <typeparam name="T">The data type to be hashed or compared by value.</typeparam>
+/// <typeparam name="TProvider">The provider of the type shape.</typeparam>
+/// <remarks>
+/// <para>
+/// The deep walking of the object graph for deep by-value equality and hashing is based on the same
+/// <see cref="GenerateShapeAttribute"/> that is used to generate MessagePack serializers.
+/// The implementation therefore considers all the same properties for equality and hashing that would
+/// be included in a serialized copy.
+/// </para>
+/// <para>
+/// This implementation is not suitable for all types. Specifically, it is not suitable for types that
+/// have multiple memory representations that are considered equal.
+/// An invariant for <see cref="IEqualityComparer{T}" /> behavior must be that if
+/// <c>x.Equals(y)</c> then <c>x.GetHashCode() == y.GetHashCode()</c>.
+/// For an auto-generated implementation of these methods for arbitrary types such as this,
+/// no specialization for multiple values that are considered equal is possible.
+/// </para>
+/// <para>
+/// For example, a <see cref="double"/> value has distinct memory representations for <c>0.0</c> and <c>-0.0</c>,
+/// yet these two values are considered equal and must have the same hash code.
+/// In this case and for several other common data types included with .NET, special consideration is built-in
+/// for correct operation.
+/// But this cannot be done automatically for any user-defined types.
+/// </para>
+/// <para>
+/// When using user-defined types for which this implementation is inappropriate,
+/// a custom implementation of <see cref="IEqualityComparer{T}"/> may be used if the type is used directly.
+/// But if the type is referenced in a type reference graph such that is used for by-value comparison,
+/// implementing <see cref="IDeepSecureEqualityComparer{T}"/> on that type will allow the type to take control
+/// of just its contribution to the hash code and equality comparison.
+/// </para>
+/// <para>
+/// Types that define no (public or opted in) properties and do not implement <see cref="IDeepSecureEqualityComparer{T}"/> will throw a <see cref="NotSupportedException"/> when attempting to create an equality comparer.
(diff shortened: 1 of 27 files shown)


src/Nerdbank.MessagePack/ByValueEqualityComparer`2.cs
```diff
@@ -0,0 +1,85 @@
...
+using Nerdbank.MessagePack.SecureHash;
+
+namespace Nerdbank.MessagePack;
+
+/// <summary>
+/// Provides deep by-value implementations of <see cref="IEqualityComparer{T}"/> for arbitrary data types.
+/// </summary>
+/// <typeparam name="T">The data type to be hashed or compared by value.</typeparam>
+/// <typeparam name="TProvider">The provider of the type shape.</typeparam>
+/// <remarks>
```
AArnott, line 13: @eiriktsarpalis If you study nothing else, I'd appreciate you studying the policies as documented in this remarks section and sending me your feedback, please.

src/Nerdbank.MessagePack/ByValueEqualityComparer`2.cs
```diff
@@ -0,0 +1,85 @@
...
+/// <para>
+/// The deep walking of the object graph for deep by-value equality and hashing is based on the same
+/// <see cref="GenerateShapeAttribute"/> that is used to generate MessagePack serializers.
+/// The implementation therefore considers all the same properties for equality and hashing that would
+/// be included in a serialized copy.
+/// </para>
+/// <para>
+/// This implementation is not suitable for all types. Specifically, it is not suitable for types that
+/// have multiple memory representations that are considered equal.
```
eiriktsarpalis, line 22: Does this mean that the comparer might not honor custom `IEquatable<T>` implementations collapsing multiple representations?

src/Nerdbank.MessagePack/SecureHash/ByValueEnumerableEqualityComparer`2.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+	/// <inheritdoc/>
+	public bool Equals(TEnumerable? x, TEnumerable? y)
+	{
+		if (x is null || y is null)
+		{
+			return ReferenceEquals(x, y);
+		}
+
+		IEnumerable<TElement> enumerableX = getEnumerable(x);
```
eiriktsarpalis, line 28: Specializing a comparer for types implementing `IList<T>` should eliminate enumerator allocations for a lot of types.

src/Nerdbank.MessagePack/ByValueEqualityComparer`2.cs
AArnott: That's right. Implementing `IEquatable<T>` alone is something some types may already do, but honoring that when the type hasn't provided an equivalent (and/or secure) GetHashCode implementation can split the behavior and violate the invariant, where two distinct value may be seen as equal, yet produce unique hash codes.
There also isn't a guarantee that `IEquatable<T>` was implemented with 'deep by value' semantics.
So I thought it best to define a new interface that guarantees that all these elements are present.

src/Nerdbank.MessagePack/SecureHash/SecureEnumerableEqualityComparer`2.cs
```diff
@@ -0,0 +1,65 @@
...
+
+		return !enumeratorY.MoveNext();
+	}
+
+	/// <inheritdoc/>
+	public override long GetSecureHashCode([DisallowNull] TEnumerable obj)
+	{
+		IEnumerable<TElement> enumerable = getEnumerable(obj);
+
+		// Ideally we could switch this to a SIP hash implementation that can process additional data in chunks with a constant amount of memory.
```
eiriktsarpalis, line 56: This does look like a fairly expensive way for combining hashes, does SipHash require that the full list of constituents must be preserved as an input?

AArnott: Are you talking about the fact that I have to allocate a list, or that I have to hash each individual member?

The allocation can theoretically be removed once we have a different SIP hash implementation.

We could also replace hashing of each element with hashing all the elements together if the following conditions hold:
1. The elements are all `unmanaged`.
2. The enumerable supports the `CollectionMarshal.GetSpan` method (or whatever it's called)
3. The element type does not implement `IDeepSecureEqualityComparer` or otherwise does not require normalization of its values.

eiriktsarpalis: > Are you talking about the fact that I have to allocate a list, or that I have to hash each individual member?

The allocation, which I presumes stems from a requirement that all constituent hashes must be passed as inputs? Could the algorithm be adapted such that a hash is combined incrementally over a fixed length of individual hashes or does that compromise its security guarantees?

AArnott: Correct. Internally, the SIP hash implementation I have just iterates over the span in 64-bit blocks though, so it seems likely that I could alter the implementation myself to allow for simply iterating over the enumerable and incrementally hashing each element without ever allocating anything to store intermediate hashes if I can't find an implementation that already does this.

AArnott: Is there any way I can create an `unmanaged` optimized implementation and actually create it from your visitor class, without using reflection? C# won't let me because the T that's passed to the visitor doesn't have that constraint.

src/Nerdbank.MessagePack/SecureHash/ByValueEnumerableEqualityComparer`2.cs
AArnott: #65

PR #63, opened 2024-11-11
Add migration doc, analyzer, and code fix to automate migration

AArnott:


test/Nerdbank.MessagePack.Analyzers.Tests/MigrationAnalyzerTests.cs
```diff
@@ -0,0 +1,147 @@
...
+		// Verify migration when the formatter accesses types and members with the old library's
+		// Extension*Format* naming pattern and update it to drop the 'Format' portion.
+	}
+
+	private Task VerifyCodeFixAsync([StringSyntax("c#-test")] string source, [StringSyntax("c#-test")] string fixedSource)
+	{
+		return new VerifyCS.Test
+		{
+			TestCode = source.Replace("\t", "    "),
+			FixedCode = fixedSource.Replace("\t", "    "),
```
AArnott, line 243: @sharwell Is there a way to configure roslyn to use tabs for formatting within this test harness?

test/Nerdbank.MessagePack.Analyzers.Tests/MigrationAnalyzerTests.cs
```diff
@@ -0,0 +1,147 @@
...
+	[Fact(Skip = "Not implemented")]
+	public void FormatterWithExtensionReferences()
+	{
+		// Verify migration when the formatter accesses types and members with the old library's
+		// Extension*Format* naming pattern and update it to drop the 'Format' portion.
+	}
+
+	private Task VerifyCodeFixAsync([StringSyntax("c#-test")] string source, [StringSyntax("c#-test")] string fixedSource)
+	{
+		return new VerifyCS.Test
```
AArnott, line 240: @sharwell How do I inject an external source generator into the test harness? The "fixed" code only compiles when combined with a source generator. Right now, that's the only thing keeping my `Formatter` test from passing.
The formatter test fails with:
> The upper limit for the number of code fix iterations was exceeded

But when I look into it, the test harness thinks there is more work to do because the source generator is missing and a compile error results. Once we fix that, the fix provider will (by design) require two iterations to complete, and I wonder if the test harness will figure out that that's acceptable automatically, since the initial version of the test code has only one diagnostic, and the fix produces code that generates a new diagnostic, which is then fixed.

test/Nerdbank.MessagePack.Analyzers.Tests/MigrationAnalyzerTests.cs
```diff
@@ -0,0 +1,147 @@
...
+			using Nerdbank.MessagePack;
+			using PolyType;
+			
+			[MessagePackConverter(typeof(MyTypeFormatter))]
+			public class MyType
+			{
+				public string Name { get; set; }
+			}
+
+			[GenerateShape<string>]
```
AArnott, line 87: @eiriktsarpalis What are the implications of potentially many user types adding such attributes like this within one project? Will `string`'s shape be described many times in the .dll, bloating its size?

eiriktsarpalis: Yes. Each witness class generates its own fully encapsulated `ITypeShapeProvider` implementation. This approach predates static abstracts and `IShapeable<T>` and is just a line-for-line copy of the STJ source generator and `JsonSerializerContext`.

It seems reasonable to investigate an alternative strategy where only one `ITypeShapeProvider` is being generated per project and individual witnesses point to that using stubs. That could create issues in the future assuming individual witnesses could configure aspects of source generation though (although I don't have any such features in mind at the moment).

AArnott: Filed https://github.com/eiriktsarpalis/PolyType/issues/47

test/Nerdbank.MessagePack.Analyzers.Tests/MigrationAnalyzerTests.cs
```diff
@@ -0,0 +1,196 @@
...
+	public async Task MessagePackObject_Keys()
+	{
+		string source = /* lang=c#-test */ """
+			using MessagePack;
+			using MessagePack.Formatters;
+
+			[{|NBMsgPack102:MessagePackObject|}]
+			public class MyType
+			{
+				[{|NBMsgPack103:Key(0)|}]
```
AArnott, line 143: @sharwell this and the `MessagePackObject_Map` test are failing because it takes 2 iterations instead of 1. 

I expect it to be able to be fixed in just 1 iteration as well, since these are two distinct diagnostic IDs, both emitted initially, and the code fix provider is written to offer code fixes all the diagnostics at once.
But your test harness only delivers one diagnostic at a time to the code fix provider. I don't know if that's why we get 2 iterations. What should I do here?

AArnott:
I'm merging the PR with the two failing tests disabled. But I'm still keenly interested in re-enabling them after I get questions answered, @sharwell. :wink:

PR #114, opened 2024-11-22
System.Numeric.Vector based Hardware Intrinsics

pCYSl5EDgo:
I do not know well how to add PolyType-based custom type converters.
This implementation is different from the one of `MessagePack-CSharp`. I'll write benchmark code which is more performant.

### DIFF
--- a/src/Nerdbank.MessagePack/Converters/HardwareAccelerated.cs
+++ b/src/Nerdbank.MessagePack/Converters/HardwareAccelerated.cs
@@ -7,7 +7,6 @@ using System.Numerics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Runtime.Intrinsics;
-using DecodeResult = Nerdbank.MessagePack.MessagePackPrimitives.DecodeResult;
 
 namespace Nerdbank.MessagePack.Converters;
 
@@ -59,7 +58,7 @@ internal static class HardwareAccelerated
 
 		if (typeof(TElement) == typeof(bool))
 		{
-			converter = new PrimitiveArrayConverter<TEnumerable, bool>(spanConstructorKind);
+			converter = new BoolArrayConverter<TEnumerable>(spanConstructorKind);
 			return true;
 		}
 		else if (typeof(TElement) == typeof(sbyte))
@@ -112,26 +111,46 @@ internal static class HardwareAccelerated
 		return false;
 	}
 
-	private static ref TElement TryGetReferenceAndLength<TEnumerable, TElement>(TEnumerable enumerable, out int length)
+	private static ref TElement GetReferenceAndLength<TEnumerable, TElement>(SpanConstructorKind spanConstructorKind, TEnumerable enumerable, out int length)
+		where TElement : unmanaged
 	{
-		switch (enumerable)
+		switch (spanConstructorKind)
 		{
-			case TElement[] array:
+			case SpanConstructorKind.Array:
+				TElement[] array = Unsafe.As<TEnumerable, TElement[]>(ref enumerable);
 				length = array.Length;
 				return ref MemoryMarshal.GetArrayDataReference(array);
-			case List<TElement> list:
+			case SpanConstructorKind.List:
+				List<TElement> list = Unsafe.As<TEnumerable, List<TElement>>(ref enumerable);
 				length = list.Count;
 				return ref MemoryMarshal.GetReference(CollectionsMarshal.AsSpan(list));
-			case ReadOnlyMemory<TElement> rom:
+			case SpanConstructorKind.ReadOnlyMemory:
+				ReadOnlyMemory<TElement> rom = Unsafe.As<TEnumerable, ReadOnlyMemory<TElement>>(ref enumerable);
 				length = rom.Length;
 				return ref MemoryMarshal.GetReference(rom.Span);
-			case Memory<TElement> mem:
+			default:
+				Memory<TElement> mem = Unsafe.As<TEnumerable, Memory<TElement>>(ref enumerable);
 				length = mem.Length;
 				return ref MemoryMarshal.GetReference(mem.Span);
 		}
+	}
 
-		Unsafe.SkipInit(out length);
-		return ref Unsafe.NullRef<TElement>();
+	private static TEnumerable GetEmptyEnumerable<TEnumerable, TElement>(SpanConstructorKind spanConstructorKind)
+		where TElement : unmanaged
+	{
+		switch (spanConstructorKind)
+		{
+			case SpanConstructorKind.Array:
+				return (TEnumerable)(object)Array.Empty<TElement>();
+			case SpanConstructorKind.List:
+				return (TEnumerable)(object)new List<TElement>();
+			case SpanConstructorKind.ReadOnlyMemory:
+				ReadOnlyMemory<TElement> rom = ReadOnlyMemory<TElement>.Empty;
+				return Unsafe.As<ReadOnlyMemory<TElement>, TEnumerable>(ref rom)!;
+			default:
+				Memory<TElement> mem = Memory<TElement>.Empty;
+				return Unsafe.As<Memory<TElement>, TEnumerable>(ref mem)!;
+		}
 	}
 
 	/// <summary>
@@ -139,376 +158,6 @@ internal static class HardwareAccelerated
 	/// </summary>
 	private static class MessagePackPrimitiveSpanUtility
 	{
-		internal static unsafe T1 Interpret<T0, T1>(ref byte input, int offset)
-			where T0 : unmanaged
-			where T1 : unmanaged
-		{
-			T0 value = sizeof(T0) == 1 ? Unsafe.BitCast<byte, T0>(Unsafe.Add(ref input, offset)) : Unsafe.ReadUnaligned<T0>(ref Unsafe.Add(ref input, offset));
-			if (BitConverter.IsLittleEndian)
-			{
-				if (typeof(T0) == typeof(short) || typeof(T0) == typeof(ushort))
-				{
-					value = Unsafe.BitCast<short, T0>(BinaryPrimitives.ReverseEndianness(Unsafe.BitCast<T0, short>(value)));
-				}
-				else if (typeof(T0) == typeof(int) || typeof(T0) == typeof(uint) || typeof(T0) == typeof(float))
-				{
-					value = Unsafe.BitCast<int, T0>(BinaryPrimitives.ReverseEndianness(Unsafe.BitCast<T0, int>(value)));
-				}
-				else if (typeof(T0) == typeof(long) || typeof(T0) == typeof(ulong) || typeof(T0) == typeof(double))
-				{
-					value = Unsafe.BitCast<long, T0>(BinaryPrimitives.ReverseEndianness(Unsafe.BitCast<T0, long>(value)));
-				}
-			}
-
-			if (typeof(T0) == typeof(T1))
-			{
-				return Unsafe.BitCast<T0, T1>(value);
-			}
-
-			if (typeof(T1) == typeof(sbyte))
-			{
-				if (typeof(T0) == typeof(short))
-				{
-					return Unsafe.BitCast<sbyte, T1>((sbyte)Unsafe.BitCast<T0, short>(value));
-				}
-				else if (typeof(T0) == typeof(int))
-				{
-					return Unsafe.BitCast<sbyte, T1>((sbyte)Unsafe.BitCast<T0, int>(value));
-				}
-				else if (typeof(T0) == typeof(long))
-				{
-					return Unsafe.BitCast<sbyte, T1>((sbyte)Unsafe.BitCast<T0, long>(value));
-				}
-				else if (typeof(T0) == typeof(byte))
-				{
-					return Unsafe.BitCast<sbyte, T1>((sbyte)Unsafe.BitCast<T0, byte>(value));
-				}
-				else if (typeof(T0) == typeof(ushort))
-				{
-					return Unsafe.BitCast<sbyte, T1>((sbyte)Unsafe.BitCast<T0, ushort>(value));
-				}
-				else if (typeof(T0) == typeof(uint))
-				{
-					return Unsafe.BitCast<sbyte, T1>((sbyte)Unsafe.BitCast<T0, uint>(value));
-				}
-				else if (typeof(T0) == typeof(ulong))
-				{
-					return Unsafe.BitCast<sbyte, T1>((sbyte)Unsafe.BitCast<T0, ulong>(value));
-				}
-			}
-			else if (typeof(T1) == typeof(short))
-			{
-				if (typeof(T0) == typeof(sbyte))
-				{
-					return Unsafe.BitCast<short, T1>(Unsafe.BitCast<T0, sbyte>(value));
-				}
-				else if (typeof(T0) == typeof(int))
-				{
-					return Unsafe.BitCast<short, T1>((short)Unsafe.BitCast<T0, int>(value));
-				}
-				else if (typeof(T0) == typeof(long))
-				{
-					return Unsafe.BitCast<short, T1>((short)Unsafe.BitCast<T0, long>(value));
-				}
-				else if (typeof(T0) == typeof(byte))
-				{
-					return Unsafe.BitCast<short, T1>(Unsafe.BitCast<T0, byte>(value));
-				}
-				else if (typeof(T0) == typeof(ushort))
-				{
-					return Unsafe.BitCast<short, T1>((short)Unsafe.BitCast<T0, ushort>(value));
-				}
-				else if (typeof(T0) == typeof(uint))
-				{
-					return Unsafe.BitCast<short, T1>((short)Unsafe.BitCast<T0, uint>(value));
-				}
-				else if (typeof(T0) == typeof(ulong))
-				{
-					return Unsafe.BitCast<short, T1>((short)Unsafe.BitCast<T0, ulong>(value));
-				}
-			}
-			else if (typeof(T1) == typeof(int))
-			{
-				if (typeof(T0) == typeof(sbyte))
-				{
-					return Unsafe.BitCast<int, T1>(Unsafe.BitCast<T0, sbyte>(value));
-				}
-				else if (typeof(T0) == typeof(short))
-				{
-					return Unsafe.BitCast<int, T1>(Unsafe.BitCast<T0, short>(value));
-				}
-				else if (typeof(T0) == typeof(long))
-				{
-					return Unsafe.BitCast<int, T1>((int)Unsafe.BitCast<T0, long>(value));
-				}
-				else if (typeof(T0) == typeof(byte))
-				{
-					return Unsafe.BitCast<int, T1>(Unsafe.BitCast<T0, byte>(value));
-				}
-				else if (typeof(T0) == typeof(ushort))
-				{
-					return Unsafe.BitCast<int, T1>(Unsafe.BitCast<T0, ushort>(value));
-				}
-				else if (typeof(T0) == typeof(uint))
-				{
-					return Unsafe.BitCast<int, T1>((int)Unsafe.BitCast<T0, uint>(value));
-				}
-				else if (typeof(T0) == typeof(ulong))
-				{
-					return Unsafe.BitCast<int, T1>((int)Unsafe.BitCast<T0, ulong>(value));
-				}
-			}
-			else if (typeof(T1) == typeof(long))
-			{
-				if (typeof(T0) == typeof(sbyte))
-				{
-					return Unsafe.BitCast<long, T1>(Unsafe.BitCast<T0, sbyte>(value));
-				}
-				else if (typeof(T0) == typeof(short))
-				{
-					return Unsafe.BitCast<long, T1>(Unsafe.BitCast<T0, short>(value));
-				}
-				else if (typeof(T0) == typeof(int))
-				{
-					return Unsafe.BitCast<long, T1>(Unsafe.BitCast<T0, int>(value));
-				}
-				else if (typeof(T0) == typeof(byte))
-				{
-					return Unsafe.BitCast<long, T1>(Unsafe.BitCast<T0, byte>(value));
-				}
-				else if (typeof(T0) == typeof(ushort))
-				{
-					return Unsafe.BitCast<long, T1>(Unsafe.BitCast<T0, ushort>(value));
-				}
-				else if (typeof(T0) == typeof(uint))
-				{
-					return Unsafe.BitCast<long, T1>(Unsafe.BitCast<T0, uint>(value));
-				}
-				else if (typeof(T0) == typeof(ulong))
-				{
-					return Unsafe.BitCast<long, T1>((long)Unsafe.BitCast<T0, ulong>(value));
-				}
-			}
-			else if (typeof(T1) == typeof(ushort))
-			{
-				if (typeof(T0) == typeof(sbyte))
-				{
-					return Unsafe.BitCast<ushort, T1>((ushort)Unsafe.BitCast<T0, sbyte>(value));
-				}
-				else if (typeof(T0) == typeof(short))
-				{
-					return Unsafe.BitCast<ushort, T1>((ushort)Unsafe.BitCast<T0, short>(value));
-				}
-				else if (typeof(T0) == typeof(int))
-				{
-					return Unsafe.BitCast<ushort, T1>((ushort)Unsafe.BitCast<T0, int>(value));
-				}
-				else if (typeof(T0) == typeof(long))
-				{
-					return Unsafe.BitCast<ushort, T1>((ushort)Unsafe.BitCast<T0, long>(value));
-				}
-				else if (typeof(T0) == typeof(byte))
-				{
-					return Unsafe.BitCast<ushort, T1>(Unsafe.BitCast<T0, byte>(value));
-				}
-				else if (typeof(T0) == typeof(uint))
-				{
-					return Unsafe.BitCast<ushort, T1>((ushort)Unsafe.BitCast<T0, uint>(value));
-				}
-				else if (typeof(T0) == typeof(ulong))
-				{
-					return Unsafe.BitCast<ushort, T1>((ushort)Unsafe.BitCast<T0, ulong>(value));
-				}
-			}
-			else if (typeof(T1) == typeof(uint))
-			{
-				if (typeof(T0) == typeof(sbyte))
-				{
-					return Unsafe.BitCast<uint, T1>((uint)Unsafe.BitCast<T0, sbyte>(value));
-				}
-				else if (typeof(T0) == typeof(short))
-				{
-					return Unsafe.BitCast<uint, T1>((uint)Unsafe.BitCast<T0, short>(value));
-				}
-				else if (typeof(T0) == typeof(int))
-				{
-					return Unsafe.BitCast<uint, T1>((uint)Unsafe.BitCast<T0, int>(value));
-				}
-				else if (typeof(T0) == typeof(long))
-				{
-					return Unsafe.BitCast<uint, T1>((uint)Unsafe.BitCast<T0, long>(value));
-				}
-				else if (typeof(T0) == typeof(byte))
-				{
-					return Unsafe.BitCast<uint, T1>(Unsafe.BitCast<T0, byte>(value));
-				}
-				else if (typeof(T0) == typeof(ushort))
-				{
-					return Unsafe.BitCast<uint, T1>(Unsafe.BitCast<T0, ushort>(value));
-				}
-				else if (typeof(T0) == typeof(ulong))
-				{
-					return Unsafe.BitCast<uint, T1>((uint)Unsafe.BitCast<T0, ulong>(value));
-				}
-			}
-			else if (typeof(T1) == typeof(ulong))
-			{
-				if (typeof(T0) == typeof(sbyte))
-				{
-					return Unsafe.BitCast<ulong, T1>((ulong)Unsafe.BitCast<T0, sbyte>(value));
-				}
-				else if (typeof(T0) == typeof(short))
-				{
-					return Unsafe.BitCast<ulong, T1>((ulong)Unsafe.BitCast<T0, short>(value));
-				}
-				else if (typeof(T0) == typeof(int))
-				{
-					return Unsafe.BitCast<ulong, T1>((ulong)Unsafe.BitCast<T0, int>(value));
-				}
-				else if (typeof(T0) == typeof(long))
-				{
-					return Unsafe.BitCast<ulong, T1>((ulong)Unsafe.BitCast<T0, long>(value));
-				}
-				else if (typeof(T0) == typeof(byte))
-				{
-					return Unsafe.BitCast<ulong, T1>(Unsafe.BitCast<T0, byte>(value));
-				}
-				else if (typeof(T0) == typeof(ushort))
-				{
-					return Unsafe.BitCast<ulong, T1>(Unsafe.BitCast<T0, ushort>(value));
-				}
-				else if (typeof(T0) == typeof(uint))
-				{
-					return Unsafe.BitCast<ulong, T1>(Unsafe.BitCast<T0, uint>(value));
-				}
-			}
-			else if (typeof(T1) == typeof(float))
-			{
-				if (typeof(T0) == typeof(sbyte))
-				{
-					return Unsafe.BitCast<float, T1>(Unsafe.BitCast<T0, sbyte>(value));
-				}
-				else if (typeof(T0) == typeof(short))
-				{
-					return Unsafe.BitCast<float, T1>(Unsafe.BitCast<T0, short>(value));
-				}
-				else if (typeof(T0) == typeof(int))
-				{
-					return Unsafe.BitCast<float, T1>(Unsafe.BitCast<T0, int>(value));
-				}
-				else if (typeof(T0) == typeof(long))
-				{
-					return Unsafe.BitCast<float, T1>(Unsafe.BitCast<T0, long>(value));
-				}
-				else if (typeof(T0) == typeof(byte))
-				{
-					return Unsafe.BitCast<float, T1>(Unsafe.BitCast<T0, byte>(value));
-				}
-				else if (typeof(T0) == typeof(ushort))
-				{
-					return Unsafe.BitCast<float, T1>(Unsafe.BitCast<T0, ushort>(value));
-				}
-				else if (typeof(T0) == typeof(uint))
-				{
-					return Unsafe.BitCast<float, T1>(Unsafe.BitCast<T0, uint>(value));
-				}
-				else if (typeof(T0) == typeof(ulong))
-				{
-					return Unsafe.BitCast<float, T1>(Unsafe.BitCast<T0, ulong>(value));
-				}
-				else if (typeof(T0) == typeof(double))
-				{
-					return Unsafe.BitCast<float, T1>((float)Unsafe.BitCast<T0, double>(value));
-				}
-			}
-			else if (typeof(T1) == typeof(double))
-			{
-				if (typeof(T0) == typeof(sbyte))
-				{
-					return Unsafe.BitCast<double, T1>(Unsafe.BitCast<T0, sbyte>(value));
-				}
-				else if (typeof(T0) == typeof(short))
-				{
-					return Unsafe.BitCast<double, T1>(Unsafe.BitCast<T0, short>(value));
-				}
-				else if (typeof(T0) == typeof(int))
-				{
-					return Unsafe.BitCast<double, T1>(Unsafe.BitCast<T0, int>(value));
-				}
-				else if (typeof(T0) == typeof(long))
-				{
-					return Unsafe.BitCast<double, T1>(Unsafe.BitCast<T0, long>(value));
-				}
-				else if (typeof(T0) == typeof(byte))
-				{
-					return Unsafe.BitCast<double, T1>(Unsafe.BitCast<T0, byte>(value));
-				}
-				else if (typeof(T0) == typeof(ushort))
-				{
-					return Unsafe.BitCast<double, T1>(Unsafe.BitCast<T0, ushort>(value));
-				}
-				else if (typeof(T0) == typeof(uint))
-				{
-					return Unsafe.BitCast<double, T1>(Unsafe.BitCast<T0, uint>(value));
-				}
-				else if (typeof(T0) == typeof(ulong))
-				{
-					return Unsafe.BitCast<double, T1>(Unsafe.BitCast<T0, ulong>(value));
-				}
-				else if (typeof(T0) == typeof(float))
-				{
-					return Unsafe.BitCast<double, T1>(Unsafe.BitCast<T0, float>(value));
-				}
-			}
-
-			throw new MessagePackSerializationException("Invalid numeric value.");
-		}
-
-		internal static DecodeResult Read<T>(ref T output, int outputCapacity, in byte msgpack, int msgpackLength, out int writtenLength, out int readLength)
-			where T : unmanaged
-		{
-			if (typeof(T) == typeof(ushort))
-			{
-				return Read(ref Unsafe.As<T, ushort>(ref output), outputCapacity, in msgpack, msgpackLength, out writtenLength, out readLength);
-			}
-			else if (typeof(T) == typeof(uint))
-			{
-				return Read(ref Unsafe.As<T, uint>(ref output), outputCapacity, in msgpack, msgpackLength, out writtenLength, out readLength);
-			}
-			else if (typeof(T) == typeof(ulong))
-			{
-				return Read(ref Unsafe.As<T, ulong>(ref output), outputCapacity, in msgpack, msgpackLength, out writtenLength, out readLength);
-			}
-			else if (typeof(T) == typeof(sbyte))
-			{
-				return Read(ref Unsafe.As<T, sbyte>(ref output), outputCapacity, in msgpack, msgpackLength, out writtenLength, out readLength);
-			}
-			else if (typeof(T) == typeof(short))
-			{
-				return Read(ref Unsafe.As<T, short>(ref output), outputCapacity, in msgpack, msgpackLength, out writtenLength, out readLength);
-			}
-			else if (typeof(T) == typeof(int))
-			{
-				return Read(ref Unsafe.As<T, int>(ref output), outputCapacity, in msgpack, msgpackLength, out writtenLength, out readLength);
-			}
-			else if (typeof(T) == typeof(long))
-			{
-				return Read(ref Unsafe.As<T, long>(ref output), outputCapacity, in msgpack, msgpackLength, out writtenLength, out readLength);
-			}
-			else if (typeof(T) == typeof(float))
-			{
-				return Read(ref Unsafe.As<T, float>(ref output), outputCapacity, in msgpack, msgpackLength, out writtenLength, out readLength);
-			}
-			else if (typeof(T) == typeof(double))
-			{
-				return Read(ref Unsafe.As<T, double>(ref output), outputCapacity, in msgpack, msgpackLength, out writtenLength, out readLength);
-			}
-			else
-			{
-				throw new NotSupportedException();
-			}
-		}
-
 		internal static nuint Write<T>(ref byte output, in T values, nuint inputLength)
 			where T : unmanaged
 		{
@@ -653,7 +302,6 @@ internal static class HardwareAccelerated
 			return inputLength;
 		}
 
-		/// <inheritdoc cref="Write(ref byte, ref bool, nuint)"/>
 		private static nuint Write(ref byte output, ref sbyte input, nuint inputLength)
 		{
 			nuint outputOffset = 0;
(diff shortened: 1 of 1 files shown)


AArnott:
I'm preparing a push to your branch to demonstrate wiring one of these up.

AArnott:
What I've pushed demonstrates basic wire-up. It only works on `T[]` and not `Memory<T>` or other variants, so I'm preparing an enhancement there, but I hope this unblocks you at least.

AArnott:
Actually, don't bother repeating the pattern I've added too much... it might change radically in the next couple hours such that it'll scale better, apply to various contiguous array types, etc.

src/Nerdbank.MessagePack/MessagePackPrimitiveSpanUtility.cs
```diff
@@ -0,0 +1,1831 @@
...
+
+		return outputOffset;
+	}
+
+	/// <summary>
+	/// Write <see cref="short"/> values.
+	/// </summary>
+	/// <param name="output">The reference to the <see cref="byte"/> buffer which receives the serialized values.</param>
+	/// <param name="input">The reference to the source <see cref="short"/> buffer.</param>
+	/// <param name="inputLength">The length of the input/output buffers.</param>
```
AArnott: The buffers are made up of elements with different individual lengths. We should be careful to document whether this is the *count* of elements, or the byte length of the buffers.

src/Nerdbank.MessagePack/MessagePackPrimitiveSpanUtility.cs
```diff
@@ -0,0 +1,1831 @@
...
+internal static class MessagePackPrimitiveSpanUtility
+{
+	/// <summary>
+	/// Read <see cref="bool"/> values.
+	/// </summary>
+	/// <param name="output">The reference to the bool buffer which receives the values.</param>
+	/// <param name="input">The reference to the source byte buffer.</param>
+	/// <param name="length">The length of the input/output buffers.</param>
+	/// <returns><see langword="true" /> if <paramref name="input"/> were all valid; otherwise, <see langword="false" />.</returns>
+	public static bool Read(ref bool output, ref byte input, nuint length)
```
AArnott: In this case, the input length can be trivially predicted because a msgpack bool value is always exactly 1 byte. But how do you deal with variable-length elements such as integers? If we can't predict how long the input buffer is before starting the read, we can't ensure they are contiguous in memory (since we deserialize from a `ReadOnlySequence<byte>`).

AArnott: I guess the answer to this is you can't. At least, [it appears you don't](https://github.com/MessagePack-CSharp/MessagePack-CSharp/blob/e6f3d81f96578595c675966747dfdd5cac26528a/src/MessagePack.Experimental/HardwareIntrinsics/FormatterHelpers.cs#L305-L325).

AArnott:
I force-pushed to erase some of my bad ideas. I think this is pretty good, although I had to comment out the `int` converter for now because it was causing tests to fail. I suspect I have bugs elsewhere, but maybe there's a bug in the accelerated converter. I'll investigate later.

This also led me to realize I don't have any tests for serializing `byte[]` and other forms of byte buffers to ensure that it optimizes for msgpack binary, and can deserialize either binary or msgpack arrays. I'm working on adding those tests (and fixing the bugs they found) which I'll merge *first* in another PR so that we can verify this PR doesn't break anything in that area.

AArnott:
It was a bug in your int encoder after all. https://github.com/AArnott/Nerdbank.MessagePack/pull/114/commits/1d0115e062bcc88383a3ed8019d2e946cf0c796c

src/Nerdbank.MessagePack/MessagePackPrimitiveSpanUtility.cs
pCYSl5EDgo: Thank you for rewriting and fixing!

src/Nerdbank.MessagePack/MessagePackPrimitiveSpanUtility.cs
pCYSl5EDgo: No I can't.

AArnott:
I added a benchmark. Somehow the accelerated code path is _worse_. Can you take a look?

| Method                      | Mean     | Error    | StdDev   | Ratio | RatioSD | Allocated | Alloc Ratio |
|---------------------------- |---------:|---------:|---------:|------:|--------:|----------:|------------:|
| Int32_Serialize             | 67.71 us | 1.033 us | 0.967 us |  2.35 |    0.06 |         - |          NA |
| Int32_Serialize_Unoptimized | 28.83 us | 0.553 us | 0.568 us |  1.00 |    0.03 |         - |          NA |

pCYSl5EDgo:
Good morning.
Um... messagepack-csharp way seems better...
Thank you for the benchmark.

pCYSl5EDgo:
Prvious impl for `MessagePack-CSharp` used `shuffle` simd api.
Current impl uses shift and or apis.
Method call count increase seems causing performance worsening.

pCYSl5EDgo:
// * Summary *

BenchmarkDotNet v0.14.0, Windows 11 (10.0.22631.4460/23H2/2023Update/SunValley3)
AMD Ryzen 7 8840U w/ Radeon 780M Graphics, 1 CPU, 16 logical and 8 physical cores
.NET SDK 9.0.100
  [Host]     : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  DefaultJob : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI

Categories=int32,serialize

| Method                      | Mean     | Error    | StdDev   | Ratio | Allocated | Alloc Ratio |
|---------------------------- |---------:|---------:|---------:|------:|----------:|------------:|
| Int32_Serialize             | 14.49 us | 0.284 us | 0.425 us |  0.41 |         - |          NA |
| Int32_Serialize_Unoptimized | 34.98 us | 0.214 us | 0.167 us |  1.00 |         - |          NA |

// * Hints *
Outliers
  HardwareAccelerated.Int32_Serialize: Default             -> 3 outliers were removed (16.21 us..16.78 us)
  HardwareAccelerated.Int32_Serialize_Unoptimized: Default -> 3 outliers were removed (35.61 us..35.93 us)

AArnott:
huh. I'll run the benchmark a few times on a few devices to see if it was a specific case that makes it slower.

AArnott:
The benchmark results I see on the one device I tried are repeatable. When I try on another machine, I get a 0.57 ratio instead of a 2.2 one. I'll run ETW on the 2.2 ratio device to see what is causing it to go so slowly.

AArnott:
Most machines I'm testing on show substantial improvement. The ETW on the oddly slower machine shows that all time is spent in the `WriteLittleEndian` method in the "optimized" case, and in the unoptimized case the largest single method call is `MessagePackWriter.Advance`, which offers an idea...
Even when hardware acceleration isn't used, if I just optimize the primitive array converters to allocate all the buffer in advance and thereby have a tighter write loop, I wonder if I can actually compete with the hardware-accelerated performance. I'm going to try it.

pCYSl5EDgo:
> Even when hardware acceleration isn't used, if I just optimize the primitive array converters to allocate all the buffer in advance and thereby have a tighter write loop, I wonder if I can actually compete with the hardware-accelerated performance. I'm going to try it.

[According to my old benchmark](https://github.com/MessagePack-CSharp/MessagePack-CSharp/pull/1744#issuecomment-1907627401), the optimization you said corresponds to .NET6. Hardware Intrinsics optimization corresponds to .NET8.
Bulk memory allocation and tighter write loop improves performance substantially.

AArnott:
Here is what I get on an ARM64 machine. Single and Double see huge improvements (I'm really curious as to why), and 64-bit integers see huge improvement as well. But the rest are worse.


| Method                       | Categories       | Mean      | Error      | StdDev     | Median    | Ratio | RatioSD | Allocated | Alloc Ratio |
|----------------------------- |----------------- |----------:|-----------:|-----------:|----------:|------:|--------:|----------:|------------:|
| Double_Serialize             | double,serialize | 19.975 us |  0.2409 us |  0.3605 us | 19.949 us |  0.45 |    0.01 |         - |          NA |
| Double_Serialize_Unoptimized | double,serialize | 44.096 us |  0.3469 us |  0.5192 us | 43.975 us |  1.00 |    0.02 |         - |          NA |
|                              |                  |           |            |            |           |       |         |           |             |
| Single_Serialize             | float,serialize  |  8.255 us |  0.0235 us |  0.0329 us |  8.249 us |  0.23 |    0.00 |         - |          NA |
| Single_Serialize_Unoptimized | float,serialize  | 35.898 us |  0.2215 us |  0.3105 us | 35.977 us |  1.00 |    0.01 |         - |          NA |
|                              |                  |           |            |            |           |       |         |           |             |
| Int16_Serialize              | int16,serialize  | 86.634 us |  0.2463 us |  0.3371 us | 86.633 us |  1.22 |    0.01 |         - |          NA |
| Int16_Serialize_Unoptimized  | int16,serialize  | 71.049 us |  0.5363 us |  0.8028 us | 71.064 us |  1.00 |    0.02 |         - |          NA |
|                              |                  |           |            |            |           |       |         |           |             |
| Int32_Serialize              | int32,serialize  | 80.325 us |  4.9753 us |  7.4468 us | 81.747 us |  1.10 |    0.11 |         - |          NA |
| Int32_Serialize_Unoptimized  | int32,serialize  | 73.125 us |  2.0557 us |  3.0132 us | 74.591 us |  1.00 |    0.06 |         - |          NA |
|                              |                  |           |            |            |           |       |         |           |             |
| Int64_Serialize              | int64,serialize  | 61.569 us | 11.5811 us | 16.9754 us | 76.639 us |  0.69 |    0.19 |         - |          NA |
| Int64_Serialize_Unoptimized  | int64,serialize  | 89.171 us |  0.3797 us |  0.5566 us | 89.260 us |  1.00 |    0.01 |         - |          NA |
|                              |                  |           |            |            |           |       |         |           |             |
| Int8_Serialize               | int8,serialize   | 56.701 us |  0.5252 us |  0.7861 us | 56.897 us |  1.07 |    0.03 |         - |          NA |
| Int8_Serialize_Unoptimized   | int8,serialize   | 53.220 us |  0.8710 us |  1.2491 us | 53.026 us |  1.00 |    0.03 |         - |          NA |
|                              |                  |           |            |            |           |       |         |           |             |
| UInt16_Serialize             | uint16,serialize | 68.960 us |  0.1300 us |  0.1906 us | 68.941 us |  1.93 |    0.04 |         - |          NA |
| UInt16_Serialize_Unoptimized | uint16,serialize | 35.655 us |  0.5106 us |  0.7485 us | 36.057 us |  1.00 |    0.03 |         - |          NA |
|                              |                  |           |            |            |           |       |         |           |             |
| UInt32_Serialize             | uint32,serialize | 48.141 us |  0.0671 us |  0.0983 us | 48.122 us |  1.05 |    0.00 |         - |          NA |
| UInt32_Serialize_Unoptimized | uint32,serialize | 45.642 us |  0.1288 us |  0.1762 us | 45.661 us |  1.00 |    0.01 |         - |          NA |
|                              |                  |           |            |            |           |       |         |           |             |
| UInt64_Serialize             | uint64,serialize | 18.565 us |  0.0661 us |  0.0989 us | 18.555 us |  0.36 |    0.00 |         - |          NA |
| UInt64_Serialize_Unoptimized | uint64,serialize | 52.195 us |  0.1203 us |  0.1686 us | 52.258 us |  1.00 |    0.00 |         - |          NA |

AArnott:
The `bool[]` benchmarks are phenomenal on a Ryzen 9:

| Method                       | Categories       | Mean        | Error     | StdDev    | Ratio | Gen0   | Allocated | Alloc Ratio |
|----------------------------- |----------------- |------------:|----------:|----------:|------:|-------:|----------:|------------:|
| Bool_Deserialize             | bool,deserialize |    608.9 ns |   9.79 ns |  23.07 ns |  0.02 | 1.1969 |   10024 B |        1.00 |
| Bool_Deserialize_Unoptimized | bool,deserialize | 37,250.5 ns | 348.18 ns | 290.75 ns |  1.00 | 1.1597 |   10024 B |        1.00 |
|                              |                  |             |           |           |       |        |           |             |
| Bool_Serialize               | bool,serialize   |    385.7 ns |   2.90 ns |   2.72 ns |  0.01 |      - |         - |          NA |
| Bool_Serialize_Unoptimized   | bool,serialize   | 37,594.6 ns | 256.20 ns | 227.11 ns |  1.00 |      - |         - |          NA |

pCYSl5EDgo:
Very interesting results in ARM.
Thanks.

AArnott:
I did a proof of concept and measured on my Ryzen 9:

Vector-accelerated vs. baseline:

| Method                      | Mean     | Error    | StdDev   | Ratio | Allocated | Alloc Ratio |
|---------------------------- |---------:|---------:|---------:|------:|----------:|------------:|
| Int16_Serialize             | 42.53 us | 0.289 us | 0.242 us |  0.53 |         - |          NA |
| Int16_Serialize_Unoptimized | 79.98 us | 0.446 us | 0.417 us |  1.00 |         - |          NA |

Vector-accelerated vs. optimized via a dedicated `short[]` converter that only calls `MessagePackWriter.GetSpan` and `Advance` once:

| Method                      | Mean     | Error    | StdDev   | Ratio | Allocated | Alloc Ratio |
|---------------------------- |---------:|---------:|---------:|------:|----------:|------------:|
| Int16_Serialize             | 41.20 us | 0.400 us | 0.374 us |  0.93 |         - |          NA |
| Int16_Serialize_Unoptimized | 44.32 us | 0.320 us | 0.299 us |  1.00 |         - |          NA |

We see the hardware acceleration benefit is very much reduced. And while I haven't tested it yet, I'm expecting it to perform reliably better on any machine, including the two I have that show the hardware accelerated versions being worse. 
I'm going to further develop this idea into Advance-optimized converters for all the primitive array/memory types. Then we can see how vector-accelerated compares to that, across machines.

AArnott:
I got Int16 to be _faster_ without vector acceleration:

| Method                      | Mean     | Error    | StdDev   | Ratio | Allocated | Alloc Ratio |
|---------------------------- |---------:|---------:|---------:|------:|----------:|------------:|
| Int16_Serialize             | 41.40 us | 0.275 us | 0.257 us |  1.03 |         - |          NA |
| Int16_Serialize_Unoptimized | 40.28 us | 0.425 us | 0.397 us |  1.00 |         - |          NA |

AArnott:
Ryzen9 table after my optimization without vectors. This doesn't include your "Add bulk Read" commit.

| Method                       | Categories       | Mean        | Error     | StdDev    | Median      | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|----------------------------- |----------------- |------------:|----------:|----------:|------------:|------:|--------:|-------:|----------:|------------:|
| Bool_Deserialize             | bool,deserialize |    676.2 ns |  23.58 ns |  69.52 ns |    635.4 ns |  0.02 |    0.00 | 1.1969 |   10024 B |        1.00 |
| Bool_Deserialize_Unoptimized | bool,deserialize | 33,002.3 ns | 592.72 ns | 554.43 ns | 33,010.6 ns |  1.00 |    0.02 | 1.1597 |   10024 B |        1.00 |
|                              |                  |             |           |           |             |       |         |        |           |             |
| Bool_Serialize               | bool,serialize   |    354.0 ns |   1.95 ns |   1.83 ns |    354.5 ns |  0.01 |    0.00 |      - |         - |          NA |
| Bool_Serialize_Unoptimized   | bool,serialize   | 34,377.5 ns | 324.15 ns | 303.21 ns | 34,351.4 ns |  1.00 |    0.01 |      - |         - |          NA |
|                              |                  |             |           |           |             |       |         |        |           |             |
| Double_Serialize             | double,serialize | 12,364.3 ns |  41.91 ns |  35.00 ns | 12,373.9 ns |  0.44 |    0.00 |      - |         - |          NA |
| Double_Serialize_Unoptimized | double,serialize | 28,332.5 ns | 127.41 ns | 119.18 ns | 28,319.6 ns |  1.00 |    0.01 |      - |         - |          NA |
|                              |                  |             |           |           |             |       |         |        |           |             |
| Single_Serialize             | float,serialize  | 14,944.4 ns |  82.12 ns |  76.82 ns | 14,934.9 ns |  0.73 |    0.01 |      - |         - |          NA |
| Single_Serialize_Unoptimized | float,serialize  | 20,411.5 ns | 197.73 ns | 175.29 ns | 20,416.3 ns |  1.00 |    0.01 |      - |         - |          NA |
|                              |                  |             |           |           |             |       |         |        |           |             |
| Int16_Serialize              | int16,serialize  | 40,864.6 ns | 317.22 ns | 296.73 ns | 40,897.5 ns |  0.97 |    0.01 |      - |         - |          NA |
| Int16_Serialize_Unoptimized  | int16,serialize  | 42,300.7 ns | 282.16 ns | 250.13 ns | 42,286.7 ns |  1.00 |    0.01 |      - |         - |          NA |
|                              |                  |             |           |           |             |       |         |        |           |             |
| Int32_Serialize              | int32,serialize  | 48,641.4 ns | 396.73 ns | 371.10 ns | 48,677.7 ns |  1.18 |    0.01 |      - |         - |          NA |
| Int32_Serialize_Unoptimized  | int32,serialize  | 41,306.8 ns | 308.93 ns | 288.98 ns | 41,268.0 ns |  1.00 |    0.01 |      - |         - |          NA |
|                              |                  |             |           |           |             |       |         |        |           |             |
| Int64_Serialize              | int64,serialize  | 49,294.6 ns | 417.73 ns | 390.75 ns | 49,169.5 ns |  1.04 |    0.01 |      - |         - |          NA |
| Int64_Serialize_Unoptimized  | int64,serialize  | 47,323.9 ns | 243.96 ns | 228.20 ns | 47,288.2 ns |  1.00 |    0.01 |      - |         - |          NA |
|                              |                  |             |           |           |             |       |         |        |           |             |
| Int8_Serialize               | int8,serialize   | 28,610.5 ns | 402.23 ns | 376.25 ns | 28,704.6 ns |  0.63 |    0.01 |      - |         - |          NA |
| Int8_Serialize_Unoptimized   | int8,serialize   | 45,496.0 ns | 563.78 ns | 527.36 ns | 45,491.1 ns |  1.00 |    0.02 |      - |         - |          NA |
|                              |                  |             |           |           |             |       |         |        |           |             |
| UInt16_Serialize             | uint16,serialize | 19,941.2 ns |  86.50 ns |  80.91 ns | 19,937.8 ns |  1.01 |    0.01 |      - |         - |          NA |
| UInt16_Serialize_Unoptimized | uint16,serialize | 19,822.6 ns | 148.16 ns | 138.59 ns | 19,811.4 ns |  1.00 |    0.01 |      - |         - |          NA |
|                              |                  |             |           |           |             |       |         |        |           |             |
| UInt32_Serialize             | uint32,serialize | 24,813.8 ns |  89.19 ns |  79.06 ns | 24,805.9 ns |  1.15 |    0.01 |      - |         - |          NA |
| UInt32_Serialize_Unoptimized | uint32,serialize | 21,624.4 ns |  83.94 ns |  70.10 ns | 21,589.1 ns |  1.00 |    0.00 |      - |         - |          NA |
|                              |                  |             |           |           |             |       |         |        |           |             |
| UInt64_Serialize             | uint64,serialize | 24,903.7 ns | 128.24 ns | 119.96 ns | 24,868.0 ns |  0.84 |    0.01 |      - |         - |          NA |
| UInt64_Serialize_Unoptimized | uint64,serialize | 29,540.4 ns | 234.06 ns | 207.49 ns | 29,539.5 ns |  1.00 |    0.01 |      - |         - |          NA |

pCYSl5EDgo:
Hmm.. Thank you for the benchmark.
Int16, Int32, Int64, UInt16, UInt32 needs improvement (or Vector based hardware-intrinsics is not necessary for them)

AArnott:
I updated my last table above after re-running with your bulk read change.

AArnott:
Here is my ARM64 machine:

```

BenchmarkDotNet v0.14.0, Windows 11 (10.0.22631.4460/23H2/2023Update/SunValley3)
Snapdragon 8cx Gen 3 3.0 GHz, 1 CPU, 8 logical and 8 physical cores
.NET SDK 9.0.100
  [Host]    : .NET 8.0.11 (8.0.1124.51707), Arm64 RyuJIT AdvSIMD
  MediumRun : .NET 8.0.11 (8.0.1124.51707), Arm64 RyuJIT AdvSIMD

Job=MediumRun  IterationCount=15  LaunchCount=2  
WarmupCount=10  

```
| Method                       | Categories       | Mean        | Error       | StdDev       | Median      | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|----------------------------- |----------------- |------------:|------------:|-------------:|------------:|------:|--------:|-------:|----------:|------------:|
| Bool_Deserialize             | bool,deserialize |  1,295.8 ns |     4.79 ns |      7.17 ns |  1,296.9 ns |  0.03 |    0.00 | 2.3918 |   10024 B |        1.00 |
| Bool_Deserialize_Unoptimized | bool,deserialize | 38,041.9 ns |   139.96 ns |    196.21 ns | 38,087.8 ns |  1.00 |    0.01 | 2.3804 |   10024 B |        1.00 |
|                              |                  |             |             |              |             |       |         |        |           |             |
| Bool_Serialize               | bool,serialize   |    554.8 ns |     2.19 ns |      3.15 ns |    554.9 ns |  0.01 |    0.00 |      - |         - |          NA |
| Bool_Serialize_Unoptimized   | bool,serialize   | 49,205.8 ns |   106.85 ns |    156.62 ns | 49,165.0 ns |  1.00 |    0.00 |      - |         - |          NA |
|                              |                  |             |             |              |             |       |         |        |           |             |
| Double_Serialize             | double,serialize | 19,074.1 ns |    66.16 ns |     96.98 ns | 19,055.4 ns |  0.59 |    0.00 |      - |         - |          NA |
| Double_Serialize_Unoptimized | double,serialize | 32,574.3 ns |    72.53 ns |    106.31 ns | 32,596.0 ns |  1.00 |    0.00 |      - |         - |          NA |
|                              |                  |             |             |              |             |       |         |        |           |             |
| Single_Serialize             | float,serialize  |  8,217.0 ns |    16.43 ns |     24.08 ns |  8,218.0 ns |  0.29 |    0.00 |      - |         - |          NA |
| Single_Serialize_Unoptimized | float,serialize  | 27,977.4 ns |    26.19 ns |     36.71 ns | 27,976.8 ns |  1.00 |    0.00 |      - |         - |          NA |
|                              |                  |             |             |              |             |       |         |        |           |             |
| Int16_Serialize              | int16,serialize  | 88,133.5 ns | 1,549.35 ns |  2,271.02 ns | 89,877.6 ns |  1.82 |    0.08 |      - |         - |          NA |
| Int16_Serialize_Unoptimized  | int16,serialize  | 48,519.6 ns | 1,131.12 ns |  1,693.00 ns | 48,167.0 ns |  1.00 |    0.05 |      - |         - |          NA |
|                              |                  |             |             |              |             |       |         |        |           |             |
| Int32_Serialize              | int32,serialize  | 71,665.3 ns |   908.42 ns |  1,359.68 ns | 71,635.7 ns |  1.53 |    0.04 |      - |         - |          NA |
| Int32_Serialize_Unoptimized  | int32,serialize  | 46,748.6 ns |   457.40 ns |    684.62 ns | 46,482.5 ns |  1.00 |    0.02 |      - |         - |          NA |
|                              |                  |             |             |              |             |       |         |        |           |             |
| Int64_Serialize              | int64,serialize  | 60,978.3 ns | 9,471.93 ns | 13,883.84 ns | 50,960.0 ns |  1.24 |    0.28 |      - |         - |          NA |
| Int64_Serialize_Unoptimized  | int64,serialize  | 49,118.9 ns |   995.72 ns |  1,490.35 ns | 48,907.0 ns |  1.00 |    0.04 |      - |         - |          NA |
|                              |                  |             |             |              |             |       |         |        |           |             |
| Int8_Serialize               | int8,serialize   | 57,239.1 ns |   597.05 ns |    893.64 ns | 57,004.5 ns |  1.49 |    0.04 |      - |         - |          NA |
| Int8_Serialize_Unoptimized   | int8,serialize   | 38,356.1 ns |   623.67 ns |    933.48 ns | 38,527.7 ns |  1.00 |    0.03 |      - |         - |          NA |
|                              |                  |             |             |              |             |       |         |        |           |             |
| UInt16_Serialize             | uint16,serialize | 68,838.0 ns |   205.73 ns |    288.40 ns | 68,776.1 ns |  2.33 |    0.04 |      - |         - |          NA |
| UInt16_Serialize_Unoptimized | uint16,serialize | 29,559.8 ns |   318.92 ns |    467.47 ns | 29,437.0 ns |  1.00 |    0.02 |      - |         - |          NA |
|                              |                  |             |             |              |             |       |         |        |           |             |
| UInt32_Serialize             | uint32,serialize | 48,007.8 ns |    41.66 ns |     61.07 ns | 48,000.5 ns |  1.70 |    0.01 |      - |         - |          NA |
| UInt32_Serialize_Unoptimized | uint32,serialize | 28,190.4 ns |   103.56 ns |    151.79 ns | 28,118.4 ns |  1.00 |    0.01 |      - |         - |          NA |
|                              |                  |             |             |              |             |       |         |        |           |             |
| UInt64_Serialize             | uint64,serialize | 18,442.7 ns |    41.77 ns |     59.90 ns | 18,442.8 ns |  0.56 |    0.00 |      - |         - |          NA |
| UInt64_Serialize_Unoptimized | uint64,serialize | 33,093.6 ns |   128.51 ns |    188.37 ns | 33,073.0 ns |  1.00 |    0.01 |      - |         - |          NA |

pCYSl5EDgo:
```

BenchmarkDotNet v0.14.0, Windows 11 (10.0.22631.4460/23H2/2023Update/SunValley3)
AMD Ryzen 7 8840U w/ Radeon 780M Graphics, 1 CPU, 16 logical and 8 physical cores
.NET SDK 9.0.100
  [Host]     : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  DefaultJob : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI


```
| Method           | Accelerated | Mean        | Error     | StdDev    | Gen0   | Allocated |
|----------------- |------------ |------------:|----------:|----------:|-------:|----------:|
| **Bool_Deserialize** | **False**       | **10,215.0 ns** |  **46.19 ns** |  **43.21 ns** | **1.1902** |   **10024 B** |
| **Bool_Deserialize** | **True**        |    **444.4 ns** |   **4.74 ns** |   **3.96 ns** | **1.1973** |   **10024 B** |
|                  |             |             |           |           |        |           |
| **Bool_Serialize**   | **False**       | **12,592.0 ns** | **154.49 ns** | **144.51 ns** |      **-** |         **-** |
| **Bool_Serialize**   | **True**        |    **248.8 ns** |   **1.74 ns** |   **1.45 ns** |      **-** |         **-** |
|                  |             |             |           |           |        |           |
| **Double_Serialize** | **False**       | **21,016.6 ns** | **255.80 ns** | **226.76 ns** |      **-** |         **-** |
| **Double_Serialize** | **True**        |  **4,260.2 ns** |  **19.03 ns** |  **14.86 ns** |      **-** |         **-** |
|                  |             |             |           |           |        |           |
| **Single_Serialize** | **False**       | **14,864.4 ns** |  **69.82 ns** |  **61.89 ns** |      **-** |         **-** |
| **Single_Serialize** | **True**        |  **4,271.5 ns** |  **35.18 ns** |  **32.91 ns** |      **-** |         **-** |
|                  |             |             |           |           |        |           |
| **Int16_Serialize**  | **False**       | **15,789.2 ns** |  **90.98 ns** |  **75.98 ns** |      **-** |         **-** |
| **Int16_Serialize**  | **True**        |  **8,517.1 ns** |  **83.32 ns** |  **69.57 ns** |      **-** |         **-** |
|                  |             |             |           |           |        |           |
| **Int32_Serialize**  | **False**       | **16,861.6 ns** |  **19.51 ns** |  **17.30 ns** |      **-** |         **-** |
| **Int32_Serialize**  | **True**        |  **9,847.9 ns** | **192.36 ns** | **236.24 ns** |      **-** |         **-** |
|                  |             |             |           |           |        |           |
| **Int64_Serialize**  | **False**       | **21,113.1 ns** |  **71.20 ns** |  **59.45 ns** |      **-** |         **-** |
| **Int64_Serialize**  | **True**        |  **9,282.4 ns** | **175.08 ns** | **155.20 ns** |      **-** |         **-** |
|                  |             |             |           |           |        |           |
| **Int8_Serialize**   | **False**       | **12,533.5 ns** |  **15.76 ns** |  **12.30 ns** |      **-** |         **-** |
| **Int8_Serialize**   | **True**        |  **7,536.0 ns** |  **69.02 ns** |  **61.19 ns** |      **-** |         **-** |
|                  |             |             |           |           |        |           |
| **UInt16_Serialize** | **False**       | **15,207.1 ns** | **270.82 ns** | **289.77 ns** |      **-** |         **-** |
| **UInt16_Serialize** | **True**        |  **6,462.6 ns** |  **38.70 ns** |  **34.31 ns** |      **-** |         **-** |
|                  |             |             |           |           |        |           |
| **UInt32_Serialize** | **False**       | **16,011.7 ns** |  **85.68 ns** |  **71.55 ns** |      **-** |         **-** |
| **UInt32_Serialize** | **True**        |  **6,389.9 ns** |  **44.76 ns** |  **41.87 ns** |      **-** |         **-** |
|                  |             |             |           |           |        |           |
| **UInt64_Serialize** | **False**       | **21,102.1 ns** |  **63.76 ns** |  **53.24 ns** |      **-** |         **-** |
| **UInt64_Serialize** | **True**        |  **6,412.3 ns** |  **33.71 ns** |  **26.32 ns** |      **-** |         **-** |

AArnott:
> Very sad result... hardware intrinsics is useless except bool... 😭

Bool is certainly the best result here, but it isn't the only good one. `float` and `double` also show great results on every machine, although I'd still like to understand _why_.
And I'd very much like to understand why we have such inconsistent results with hardware intrinsics across hardware. On your Ryzen 7 everything looks benefited from your accelerated code.

AArnott:
In fact with your latest version, even my ARM64 device shows the accelerated path is faster:

```

BenchmarkDotNet v0.14.0, Windows 11 (10.0.22631.4460/23H2/2023Update/SunValley3)
Snapdragon 8cx Gen 3 3.0 GHz, 1 CPU, 8 logical and 8 physical cores
.NET SDK 9.0.100
  [Host]   : .NET 8.0.11 (8.0.1124.51707), Arm64 RyuJIT AdvSIMD
  ShortRun : .NET 8.0.11 (8.0.1124.51707), Arm64 RyuJIT AdvSIMD

Job=ShortRun  IterationCount=3  LaunchCount=1  
WarmupCount=3  

```
| Method           | Accelerated | Mean        | Error       | StdDev    | Gen0   | Allocated |
|----------------- |------------ |------------:|------------:|----------:|-------:|----------:|
| **Bool_Deserialize** | **False**       | **39,546.5 ns** | **3,208.83 ns** | **175.89 ns** | **2.3804** |   **10024 B** |
| **Bool_Deserialize** | **True**        |  **1,213.8 ns** |   **145.95 ns** |   **8.00 ns** | **2.3918** |   **10024 B** |
|                  |             |             |             |           |        |           |
| **Bool_Serialize**   | **False**       | **49,348.7 ns** | **7,623.70 ns** | **417.88 ns** |      **-** |         **-** |
| **Bool_Serialize**   | **True**        |    **543.5 ns** |    **51.59 ns** |   **2.83 ns** |      **-** |         **-** |
|                  |             |             |             |           |        |           |
| **Double_Serialize** | **False**       | **32,541.0 ns** | **3,587.45 ns** | **196.64 ns** |      **-** |         **-** |
| **Double_Serialize** | **True**        |  **9,227.0 ns** |   **110.31 ns** |   **6.05 ns** |      **-** |         **-** |
|                  |             |             |             |           |        |           |
| **Single_Serialize** | **False**       | **27,836.4 ns** |   **204.76 ns** |  **11.22 ns** |      **-** |         **-** |
| **Single_Serialize** | **True**        |  **9,129.4 ns** |   **370.22 ns** |  **20.29 ns** |      **-** |         **-** |
|                  |             |             |             |           |        |           |
| **Int16_Serialize**  | **False**       | **48,236.1 ns** | **3,903.58 ns** | **213.97 ns** |      **-** |         **-** |
| **Int16_Serialize**  | **True**        | **26,563.2 ns** | **1,927.73 ns** | **105.67 ns** |      **-** |         **-** |
|                  |             |             |             |           |        |           |
| **Int32_Serialize**  | **False**       | **47,092.6 ns** | **7,754.43 ns** | **425.05 ns** |      **-** |         **-** |
| **Int32_Serialize**  | **True**        | **27,947.2 ns** | **9,953.94 ns** | **545.61 ns** |      **-** |         **-** |
|                  |             |             |             |           |        |           |
| **Int64_Serialize**  | **False**       | **41,675.2 ns** | **9,196.93 ns** | **504.12 ns** |      **-** |         **-** |
| **Int64_Serialize**  | **True**        | **34,500.9 ns** | **9,631.78 ns** | **527.95 ns** |      **-** |         **-** |
|                  |             |             |             |           |        |           |
| **Int8_Serialize**   | **False**       | **37,379.5 ns** | **7,396.86 ns** | **405.45 ns** |      **-** |         **-** |
| **Int8_Serialize**   | **True**        | **15,227.8 ns** | **4,455.14 ns** | **244.20 ns** |      **-** |         **-** |
|                  |             |             |             |           |        |           |
| **UInt16_Serialize** | **False**       | **29,004.4 ns** | **1,628.84 ns** |  **89.28 ns** |      **-** |         **-** |
| **UInt16_Serialize** | **True**        | **12,379.3 ns** |   **739.22 ns** |  **40.52 ns** |      **-** |         **-** |
|                  |             |             |             |           |        |           |
| **UInt32_Serialize** | **False**       | **28,015.6 ns** |    **46.56 ns** |   **2.55 ns** |      **-** |         **-** |
| **UInt32_Serialize** | **True**        | **12,637.1 ns** |   **599.33 ns** |  **32.85 ns** |      **-** |         **-** |
|                  |             |             |             |           |        |           |
| **UInt64_Serialize** | **False**       | **32,878.5 ns** |   **900.66 ns** |  **49.37 ns** |      **-** |         **-** |
| **UInt64_Serialize** | **True**        | **13,017.8 ns** |   **233.56 ns** |  **12.80 ns** |      **-** |         **-** |

pCYSl5EDgo:
> Ok, I see you took the vector code out of everything except bool. Makes sense I guess, considering how fast you got it without that, and your latest results are consistent across all machines.

Yes, I removed Vector-accelerated code except bool.
The method `Vector.GetElement` is slow in Ryzen. It took over 90% of execution time in Ryzen.
I heard that handling of this method is improved as Ryzen's Zen architecture evolves.
I heard that `Vector.GetElement` is *very very very* slow in ARM processor because ARM does not provide it as hardware-intrinsics and it is emulated by software.

I am now struggling in avoiding `Vector.GetElement`. Writing code which does not use `Vector.GetElement` would need some time (and there is no guarantee for performance improvements), then that code should be in another Pull Request. I would never include it in this PR.

>  I wonder why they are so slow.
 - `abstrac method`
 - Confirmation of `destination.Length` in each `TryWrite` method
  - The above causes unnecessary branching
 - Writing to `destination[0]` in each `TryWrite` method
  - The above index access causes runtime range check.
 - Assigning to `out int bytesWritten` in each `TryWrite` method
  - `out` argument is actually `pointer`. `pointer` dereference is not performant.
 - `WriteBigEndian` does not use `BinaryPrimitives.ReverseEndianness`, which is hardware-accelerated.
 - `WriteBigEndian(destination.Slice(1), *(int*)&value);`
  - `Unsafe.BitCast` is performant.

> Anyway, I wonder if we can or should 'distribute' these wins across the MessagePackWriter / MessagePackReader structs (or perhaps the MessagePackPrimitives type so that they can apply to more than just the array cases. Do you think that would be worthwhile?
Yes. That is very beneficial.

AArnott:
@pCYSl5EDgo Thank you for your continued work on this. I'm traveling today and won't likely respond to anything.

src/Nerdbank.MessagePack/Converters/HardwareAccelerated.cs
```diff
@@ -0,0 +1,2634 @@
...
+			converter = new PrimitiveArrayConverter<TEnumerable, float>(spanConstructorKind);
+			return true;
+		}
+		else if (typeof(TElement) == typeof(double))
+		{
+			converter = new PrimitiveArrayConverter<TEnumerable, double>(spanConstructorKind);
+			return true;
+		}
+
+FAIL:
```
AArnott: This label is used in just one place. Why not return from there directly?

src/Nerdbank.MessagePack/Converters/HardwareAccelerated.cs
```diff
@@ -0,0 +1,2634 @@
...
+						if (readLength >= inputLength)
+						{
+							--readLength;
+							return DecodeResult.InsufficientBuffer;
+						}
+
+						Unsafe.Add(ref output, writtenLength++) = Interpret<byte, ushort>(ref input, readLength);
+						readLength += sizeof(byte);
+						break;
+					case MessagePackCode.UInt16:
```
AArnott: There is a very large amount of code in this file. When I ran the benchmarks on my Ryzen 9 machine, deserialization was not improved for the integer types except for `Int8`. If you see the same, I wonder if we can drop a lot of the code that is redundant with our more conventional read converters for integers except for Int8. I'm feeling like "less is more" when it comes to code because it's easier to test and maintain. So we should keep it only if it's substantially better (or if we have reason to believe future chips will make it better and we want to keep the work you've done in the meantime).
BTW with regard to `Int8[]` (aka `sbyte[]`) I'm thinking we should just encode that using msgpack binary, the same as we do for the unsigned `byte[]`. I see very little lost, and what's lost is only if the deserializer doesn't even know the type they are expecting to deserialize into. But they'll gain an enormous perf boost in the process.

Thoughts?

src/Nerdbank.MessagePack/Converters/HardwareAccelerated.cs
```diff
@@ -0,0 +1,2634 @@
...
+						Unsafe.WriteUnaligned(ref Unsafe.Add(ref output, outputOffset + 1U), BinaryPrimitives.ReverseEndianness(temp));
+						outputOffset += 9U;
+						break;
+				}
+			}
+
+			return outputOffset;
+		}
+
+		private static nuint WriteBigEndian(ref byte output, ref long input, nuint inputLength)
```
AArnott: There seems to be a great deal of code duplicated between the `WriteBigEndian` and `WriteLittleEndian` method groups. In my testing today it seems that just putting a `BitConverter.IsLittleEndian` branch in a single shared method allows much more code reuse and has 0 runtime perf impact because the JIT completely removes the N/A branch of code.
Can we do that here?

pCYSl5EDgo:
<details><summary>Benchmark results for Serialize after MessagePackPrimitives improvement.</summary>
<div>

| Method             | Accelerated | Mean        | Error       | StdDev      | Median      | Gen0   | Allocated |
|------------------- |------------ |------------:|------------:|------------:|------------:|-------:|----------:|
| Double_Serialize   | False       | 11,068.6 ns |   212.53 ns |   268.78 ns | 10,961.1 ns |      - |         - |
| Double_Serialize   | True        |  4,443.1 ns |    85.31 ns |   113.89 ns |  4,476.4 ns |      - |         - |
| Single_Serialize   | False       | 10,758.0 ns |   160.50 ns |   150.13 ns | 10,810.5 ns |      - |         - |
| Single_Serialize   | True        |  4,328.1 ns |    41.26 ns |    36.57 ns |  4,334.5 ns |      - |         - |
| Int16_Serialize    | False       | 14,331.3 ns |   275.64 ns |   338.51 ns | 14,240.9 ns |      - |         - |
| Int16_Serialize    | True        | 10,129.6 ns |   201.03 ns |   396.82 ns | 10,044.8 ns |      - |         - |
| Int32_Serialize    | False       | 13,009.4 ns |   148.09 ns |   131.28 ns | 12,992.2 ns |      - |         - |
| Int32_Serialize    | True        | 10,197.9 ns |   201.94 ns |   224.46 ns | 10,085.8 ns |      - |         - |
| Int64_Serialize    | False       | 14,689.9 ns |   283.21 ns |   347.81 ns | 14,700.2 ns |      - |         - |
| Int64_Serialize    | True        |  9,401.3 ns |   145.82 ns |   136.40 ns |  9,365.8 ns |      - |         - |
| Int8_Serialize     | False       | 12,305.0 ns |   131.89 ns |   123.37 ns | 12,243.2 ns |      - |         - |
| Int8_Serialize     | True        |  6,530.0 ns |    75.54 ns |    70.66 ns |  6,504.1 ns |      - |         - |
| UInt16_Serialize   | False       | 13,064.9 ns |    93.69 ns |    83.06 ns | 13,059.4 ns |      - |         - |
| UInt16_Serialize   | True        |  6,625.7 ns |    55.68 ns |    52.08 ns |  6,646.3 ns |      - |         - |
| UInt32_Serialize   | False       | 11,999.9 ns |   162.25 ns |   143.83 ns | 11,958.9 ns |      - |         - |
| UInt32_Serialize   | True        |  6,447.0 ns |    38.31 ns |    33.96 ns |  6,434.4 ns |      - |         - |
| UInt64_Serialize   | False       | 13,081.3 ns |   105.54 ns |    88.13 ns | 13,084.6 ns |      - |         - |
| UInt64_Serialize   | True        |  6,823.5 ns |   132.98 ns |   182.02 ns |  6,797.3 ns |      - |         - |

</div></details>

src/Nerdbank.MessagePack/Converters/HardwareAccelerated.cs
pCYSl5EDgo: Ok. I rewrite it for the sake of simpleness.

src/Nerdbank.MessagePack/Converters/HardwareAccelerated.cs
pCYSl5EDgo: I do not like duplicated code but I like early-return.
I tend to use goto statement in order to go to default failure code path.
Ok, I remove this goto statement.

src/Nerdbank.MessagePack/Converters/HardwareAccelerated.cs
pCYSl5EDgo: Treating `sbyte[]` like `byte[]` will introduce a lot of performance improvement.
Interoperability between languages will decrease. If there is going to be a problem, it will be with ASP.NET.
I do not have objection due to the benefit.

I rewrote the deserialize methods as very simple codes but their perforamnces do not improve. On my machine, they are currently as twice as slower than not accelerated ones.
Next commit of mine takes some time.

AArnott: > Interoperability between languages will decrease. If there is going to be a problem, it will be with ASP.NET.

Can you elaborate on that? I want to make sure I'm not missing something. I'm thinking that deserializing a payload from another language should be just fine because C# can just deserialize as `byte[]` if that's what the other language intended.
I guess _serializing_ an `sbyte[]` may lead to other libraries/platforms misinterpreting it as `byte[]` leading to the signed integers being misinterpreted as unsigned. I guess if that's a concern we could add an option to switch back to serializing as an array of sbyte msgpack values.
What would really help is to know what the use case is for serializing an `sbyte[] `in the first place is, because I personally never used `sbyte`. Is this interoperability trade off for perf likely to pay off more than it hurts?

AArnott:
@pCYSl5EDgo FYI I'm experimenting with [some significant changes to `MessagePackReader`](https://github.com/AArnott/Nerdbank.MessagePack/compare/main...asyncreader) to make it streamable (synchronously or asynchronously). As the change is quite significant (to a few files) I wanted to call your attention to it in case the scope of your PR were to expand, I don't want it to expand into the files changes in my branch to avoid merge conflicts without us having a discussion first.

src/Nerdbank.MessagePack/Converters/HardwareAccelerated.cs
pCYSl5EDgo: I agree that using `sbyte[]` is very rare.

```csharp
enum Relationship : sbyte
{
  Banned = -1, // Comparing with 0 is efficient than with not 0 number.
  None = 0,
  Friendship = 1,
}
```

The above uses `enum` backed by `sbyte` (I sometimes write this kind of codes). If one uses the programming style of [Structure of Array](https://en.wikipedia.org/wiki/AOS_and_SOA), then they uses `sbyte[]`.
The game programmers sometimes adopt the above `SoA` programming style (See: [Unity Entity Component System](https://docs.unity3d.com/Packages/com.unity.entities@1.0/manual/index.html)).
Yeah this combination of `sbyte` and `SoA` is very very rare and will not occur for most programmers.
Those who use `sbyte[]` in cross-language env know what they are doing. It is enough to give them an option to switch back to `sbyte array` and warnings.

pCYSl5EDgo:
Today I write the code without `Vector.GetElement`, which fasten `double` serialization more than 3 times faster.
This would be in another PR after this would be merged.

# Before

| Method           | Accelerated | Mean      | Error     | StdDev    | Allocated |
|----------------- |------------ |----------:|----------:|----------:|----------:|
| Double_Serialize | False       | 11.541 μs | 0.0699 μs | 0.0620 μs |         - |
| Double_Serialize | True        |  4.536 μs | 0.0713 μs | 0.0667 μs |         - |

# After

| Method           | Accelerated | Mean      | Error     | StdDev    | Allocated |
|----------------- |------------ |----------:|----------:|----------:|----------:|
| Double_Serialize | False       | 11.300 μs | 0.2057 μs | 0.2201 μs |         - |
| Double_Serialize | True        |  1.442 μs | 0.0274 μs | 0.0256 μs |         - |

PR #117, opened 2024-11-24
Bump PolyType to 0.20.1

AArnott:
As part of this, adopt the new utilities APIs to simplify our own code.


src/Nerdbank.MessagePack/MessagePackSerializer.cs
```diff
@@ -152,29 +114,55 @@ public partial record MessagePackSerializer
...
 				lock (this.lazyInitCookie)
 				{
-					this.cachedConverters ??= this.PreserveReferences ? new(PrimitiveReferencePreservingConverters) : new(PrimitiveConverters);
+					this.cachedConverters ??= new()
+					{
+						DelayedValueFactory = new DelayedConverterFactory(),
+						ValueBuilderFactory = ctx =>
+						{
+							StandardVisitor standardVisitor = new StandardVisitor(this, ctx);
+							if (!this.PreserveReferences)
```
eiriktsarpalis, line 131: I'm guessing this is what motivates maintaining a separate cache per serializer instance? It could end up very expensive if users accidentally create new serializer instances per operation (this is a known footgun in the case of System.Text.Json).

My recommendation would be to promote the cache to a global singleton, and if an option like `PreserveReferences` is specified have that passed to the already materialized converters as part of the serialization state.

eiriktsarpalis (approved):
LGTM

src/Nerdbank.MessagePack/MessagePackSerializer.cs
AArnott: This isn't the only option on `MessagePackSerializer` that influences how the converters are initialized. [There are several](https://aarnott.github.io/Nerdbank.MessagePack/api/Nerdbank.MessagePack.MessagePackSerializer.html#properties). 
I've documented that folks should create and reuse a single instance to keep from repeating this initialization cost.

I've been thinking of a couple other solutions to this. I could put the options all in a dedicated `record` and then have a static cache of converters keyed by the options record, such that if folks recreate the same options, they'll benefit from a 'cache hit' and share the same converters. One complication to this plan is the options would necessarily include whatever custom converters the user has registered at runtime, and those are _instances_ which would need to implement by-value equality checks to be allowed to match in the cache. If they don't match, then that cache may grow arbitrarily large as the program keeps creating a new serializer with a new custom converter instance each time.

> if an option like PreserveReferences is specified have that passed to the already materialized converters as part of the serialization state.

That means each and every converter would have to be written to know about that option and honor it. Or that the converters cannot be instantiated with a pre-programmed knowledge of the sub-converters they may call, putting us back into MessagePack-CSharp design territory where it's a runtime call to ask "which converter do I use for this value" for every serialization. But we get a significant perf win over MessagePack-CSharp by _not_ making that call every time. In fact we'd be _worse_ off if we bring that back, because it _would_ be a dictionary lookup instead of MPC#'s generic static field lookup which is quite fast.

I also am concerned about a shared static cache because it could defeat garbage collection of an ALC once a type within that ALC is involved in serialization. Is there a way to avoid that problem?

And I generally hate mutable statics because of how it impacts testability, concurrency, etc.

Related to this design problem is simply the question as to whether I should expose static serialize methods. Most other serializers do. The requirement to instantiate a class first is an extra requirement, but it provides the place to store a non-static cache. If I go with the "Options" dedicated record, I could use all static methods that take an optional Options. The options object could either itself store the cache, or serve as a key to the global static cache.

MPC# not only does the static method with optional Options parameter thing (it has nothing to do with the cache though). But MPC# also has a publicly mutable static for DefaultOptions, which I absolutely despise, because most users don't supply Options, leaving them vulnerable to whatever behavior changes the rest of the app makes to that mutable static (which multiple points in the program may set and expect to hold). I've never liked that if a program/library/feature needs to set options (or defend itself against someone else changing DefaultOptions) that they have to carefully remember to supply the options parameter on every call to the serializer. I had to write an analyzer in MPC# to call out when they forget to supply the options arg. So I was trying to avoid repeating parts of that error in this new library by just building the options into the serializer itself so folks can't forget and will be more inclined to keep the serializer itself.

eiriktsarpalis: Absolutely agree that if the user can customize at runtime what converters come into play for a given type graph, then using a global static cache is a recipe for disaster. 

While creating separate caches per serializer instance is great from a state hygiene point of view, it does create the potential for degraded performance particularly in the context of a developer community that has been primed to create new instances of everything for each serialization operation. Many users of System.Text.Json have been bitten particularly badly because of this, to the extent that we had to [ship an analyzer](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1869) that detects ephemeral cache creation.

From my perspective, this conclusion has made me double down on the idea that the contract of a type should only be configurable statically, meaning that it should derive solely from the shape of type and any attribute annotations it might contains.

AArnott: > the contract of a type should only be configurable statically, meaning that it should derive solely from the shape of type and any attribute annotations it might contains.

I can see the appeal of that, for sure. Though if you allow runtime registration of custom converters at all (as opposed to requiring attributes on the data types themselves to point to the custom converter), then you've already crossed the line of having entirely statically determined serialization behavior. Doesn't STJ allow runtime registration of converters? If not, how do you handle users who need to provide a custom converter to a 3rd party type that didn't bother providing one?

eiriktsarpalis: STJ does allow runtime registration of converters, but it does occur to me that witness types today (and C# extensions in the future) potentially offer an appealing alternative for enabling this type of extensibility without the problems that runtime registration creates.

Granted, I'm not sure what this might look in the context of general-purpose library like PolyType, but it's conceivable that you could include additional metadata in `[GenerateShape<T>]` such as a type implementating a converter for `T`.

eiriktsarpalis: This is a bit of a tangent, but one potential way of addressing the problem of type customization in the general case is through type conversions. In essence you customize the shape of type A by modelling a proxy type B that has the desired contract (or pick one that already exists) and build a type converter between the two. This should let the source generator produce a shape for A that transparently reflects the structure of B.

AArnott: > witness types today (and C# extensions in the future) potentially offer an appealing alternative for enabling this type of extensibility without the problems that runtime registration creates.

True. That _is_ interesting.

> one potential way of addressing the problem of type customization in the general case is through type conversions

It's interesting, because I was thinking about that sort of option just this morning. A struct that represents just the elements that should be serialized, and has the methods to construct the original type based on those values or to construct itself based on an instance of the original type may be more intuitive for people to build than to write a custom converter.

eiriktsarpalis: Exactly. Technically it does introduce more overhead compared to writing a bespoke converter but at the end of day it isn't substantial and the approach is much more user friendly and generalized.

AArnott: The surrogate type approach is less efficient than a custom converter could theoretically be, but message pack converters that are well written are much more efficient than a handwritten converter tends to be so a surrogate approach as you're describing could be considerably better than the average converter.

PR #140, opened 2024-12-03
SIMD acceleration for float/double[]

pCYSl5EDgo:
`float[]` / `double[]` serialization is relatively easy because the ratio of output byte count by input byte count is always constant.

The `const byte B = byte.MaxValue;` and `const byte F` is introduced for the sake of the readability of very long constant binaries.

Prev: #114


AArnott (approved):
Amazing! 👏

PR #159, opened 2024-12-11
Add `MessagePackSerializer.MaxAsyncBuffer` to speed up small async deserializations

### LINKED ISSUE #155: Buffer at least some threshold of bytes before proceeding with async deserialization
As a major async perf optimization, buffer at least *n* bytes, and if EOF is encountered before reaching that threshold, just deserialize synchronously.

Async *during* deserialization is only useful for limiting memory consumption when the data to deserialize is quite large. When it is reasonably small, buffering it up front and then deserializing with the synchronous methods is much faster.

AArnott:
Since synchronous deserialization is substantially faster than async deserialization, we prefer sync. But sync requires that all msgpack data be pre-buffered. That's a reasonable trade-off, assuming the msgpack data is reasonably small. When it is large, the slower async path may be preferable to avoid unbounded memory consumption.

Closes #155

## Before

| Method            | Mean     | Error    | StdDev  | Ratio | RatioSD | Gen0    | Gen1   | Allocated | Alloc Ratio |
|------------------ |---------:|---------:|--------:|------:|--------:|--------:|-------:|----------:|------------:|
| Sync_Deserialize  | 111.4 us | 19.72 us | 1.08 us |  1.00 |    0.01 | 11.4746 | 1.9531 |  70.34 KB |        1.00 |
| Async_Deserialize | 327.0 us |  6.20 us | 0.34 us |  2.94 |    0.02 | 11.2305 | 1.9531 |  70.49 KB |        1.00 |

## After

Method | Categories | Mean | Error | StdDev | Ratio | Gen0 | Gen1 | Allocated | Alloc Ratio
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
Sync_Deserialize | map-init,Deserialize | 114.37 μs | 10.251 μs | 0.562 μs | 1.00 | 11.4746 | 1.9531 | 72024 B | 1.00
Async_Deserialize | map-init,Deserialize | 104.00 μs | 2.420 μs | 0.133 μs | 0.91 | 11.4746 | 1.9531 | 72080 B | 1.00

Not all benchmarks show 0 tax for async deserialization for small jobs. The above works because it's a reasonably _large_ array but still under the threshold for synchronous deserialization. But very tiny deserializations still pay a tax due to buffering overhead (even if it's already in memory) that sync deserialization APIs can avoid.

### DIFF
--- a/src/Nerdbank.MessagePack/MessagePackSerializer.cs
+++ b/src/Nerdbank.MessagePack/MessagePackSerializer.cs
@@ -36,6 +36,7 @@ public partial record MessagePackSerializer
 
 	private MultiProviderTypeCache? cachedConverters;
 	private bool preserveReferences;
+	private int maxAsyncBuffer = 1 * 1024 * 1024;
 
 #if NET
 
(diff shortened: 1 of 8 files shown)


AArnott:
CC: @eiriktsarpalis

PR #169, opened 2024-12-15
Add converter for `System.Drawing.Color`

### LINKED ISSUE #168: Incorrect deserialization of System.Drawing.Color property
**Bug description**
When serializing an object containing a property of type System.Drawing.Color and then deserializing that object, the value of the Color property becomes [Empty]. It is not clear whether the problem is related to the serialization or deserialization process.

**Repro steps**
C# code to reproduce, the first test fails, but the second passes (even though Point is also a struct and from the same namespace)

```C#
[GenerateShape]
public partial record Person(string Name, System.Drawing.Color ShirtColor);

[GenerateShape]
public partial record PersonWithPosition(string Name, System.Drawing.Point Position);

public class SerializerTest
{
    [Fact]
    public void SerializeAndDeserializeObjectWithColorAsProperty_ShouldPersistDataCorrectly()
    {
        // Arrange
        var serializer = new MessagePackSerializer();
        var person = new Person("John", System.Drawing.Color.MediumPurple);

        // Act
        var data = serializer.Serialize(person);
        var deserialized = serializer.Deserialize<Person>(data);

        // Assert
        Assert.Equal(person.ShirtColor, deserialized.ShirtColor);
    }

    [Fact]
    public void SerializeAndDeserializeObjectWithPointAsProperty_ShouldPersistDataCorrectly()
    {
        // Arrange
        var serializer = new MessagePackSerializer();
        var person = new PersonWithPosition("John", new System.Drawing.Point(1,1));

        // Act
        var data = serializer.Serialize(person);
        var deserialized = serializer.Deserialize<PersonWithPosition>(data);

        // Assert
        Assert.Equal(person.Position, deserialized.Position);
    }
}
```

**Expected behavior**
The value of the Color property should be the same after deserialization as it was before serialization.

**Actual behavior**
The value of the Color property becomes [Empty] after deserialization.

- Version: 0.3.38-beta
- Target Framework: .NET 8.0
- System: Windows 10

**Additional context**
Message from first test:
Assert.Equal() Failure: Values differ
Expected: Color [MediumPurple]
Actual:   Color [Empty]

AArnott:
Fixes #168

### DIFF
--- a/src/Nerdbank.MessagePack/StandardVisitor.cs
+++ b/src/Nerdbank.MessagePack/StandardVisitor.cs
@@ -4,6 +4,7 @@
 #pragma warning disable NBMsgPackAsync
 
 using System.Collections.Frozen;
+using System.Drawing;
 using System.Numerics;
 using System.Reflection;
 using System.Runtime.CompilerServices;
@@ -44,6 +45,7 @@ internal class StandardVisitor : TypeShapeVisitor, ITypeShapeFunc
 		{ typeof(byte[]), ByteArrayConverter.Instance },
 		{ typeof(Memory<byte>), new MemoryOfByteConverter() },
 		{ typeof(ReadOnlyMemory<byte>), new ReadOnlyMemoryOfByteConverter() },
+		{ typeof(Color), new SystemDrawingColorConverter() },
 #if NET
 		{ typeof(Rune), new RuneConverter() },
 		{ typeof(Int128), new Int128Converter() },
--- a/src/Nerdbank.MessagePack/Converters/BclConverters.cs
+++ b/src/Nerdbank.MessagePack/Converters/BclConverters.cs
@@ -1,0 +1,27 @@
+﻿// Copyright (c) Andrew Arnott. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#pragma warning disable SA1600 // Elements should be documented
+#pragma warning disable SA1649 // File name should match first type name
+
+using System.Drawing;
+using System.Text.Json.Nodes;
+
+namespace Nerdbank.MessagePack.Converters;
+
+internal class SystemDrawingColorConverter : MessagePackConverter<Color>
+{
+	public override Color Read(ref MessagePackReader reader, SerializationContext context)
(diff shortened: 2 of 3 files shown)


src/Nerdbank.MessagePack/StandardVisitor.cs
```diff
@@ -44,6 +45,7 @@ internal class StandardVisitor : TypeShapeVisitor, ITypeShapeFunc
 		{ typeof(byte[]), ByteArrayConverter.Instance },
 		{ typeof(Memory<byte>), new MemoryOfByteConverter() },
 		{ typeof(ReadOnlyMemory<byte>), new ReadOnlyMemoryOfByteConverter() },
+		{ typeof(Color), new SystemDrawingColorConverter() },
```
eiriktsarpalis, line 48: The downside of doing this is that the library is now always roots System.Drawing components, which should negatively impact trimmed application size. We might want to investigate better techniques that help the linker do its job properly. One possibility that I haven't verified is replacing the dictionary of base converter types with a function like so:
```C#
static Converter<T>? GetBaseCaseConverter<T>()
{
    if (typeof(T) == typeof(System.Drawing.Color)
    {
         return (Converter<T>)(object)new SystemDrawingColorConverter();
    }

    /* more base case converter types */

    return null;
}
```
_Assuming_ the linker can eliminate branches for the `T` types it knows are not being introduced, then it would never be included unless opted in by the user.

cc @eerhardt

AArnott: Yes, I had the same thought. And @pCYSl5EDgo recently contributed [a very long method](https://github.com/AArnott/Nerdbank.MessagePack/blob/6f2494b11afb0e6d8c49c36600328d18e872c8a9/src/Nerdbank.MessagePack/Converters/HardwareAccelerated.cs#L37) of type checks that I verified the JIT does in fact shorten down to just one branch. Those were all value-types however, which the JIT had to compile the method specially for each one. 

If a bunch of ref types were added, I imagine the JIT would preserve each and every ref type check in the same generated method. It also means (in the non-AOT case) that startup may suffer because we're JIT'ing more methods than before.

PR #170, opened 2024-12-15
Test on .NET 9 and test `OrderedDictionary<K, V>`

AArnott:


eiriktsarpalis:
Which makes me think I should probably also test the new collection in PolyType proper. Good to see it's working OOTB though.

PR #180, opened 2024-12-19
Make a start at adding non-generic serialize overloads

trippwill:


trippwill:
@AArnott - tried making a start on adding non-generic, Type accepting overloads.

AArnott:
Oh, I just noticed this after creating #181. 
I see you added more overloads than I did. Which ones would you directly use? We already have a _lot_ of serialize/deserialize overloads so I'm looking to keep this hopefully rare case cornered off to just a few overloads even if they're a little harder to use.

trippwill:
Looks like the primary synchronous overloads which take a shape provider and a Type instance would be the ones I would use for vs-streamjsonrpc.

I'll close this PR.

AArnott:
> take a shape provider and a Type instance

Would taking an `ITypeShape` instead (which encompasses both the Type and the shape provider) suffice?

PR #181, opened 2024-12-19
Add non-generic methods for serializing/deserializing

AArnott:


trippwill (commented):
Ignore what I said on the comment closing my PR.

I'm actually looking to use non-generics inside a custom converter. According to the analyzers, I should prefer calling context.GetConverter. Is that supportable?

AArnott:
> I'm actually looking to use non-generics inside a custom converter. According to the analyzers, I should prefer calling context.GetConverter. Is that supportable?

Ah yes, that's a different use case to be sure. And I'm glad the analyzers stopped you from making the top-level call. Let me look more into it.

trippwill:
For the vs-streamjsonrpc formatter, I'm considering an alternative: making two MessagePackSerializer instances, one for the message envelopes, and one for user data. Would it make more sense for the user to provide a shape provider (my current thinking) that gets passed to call to Serialize and GetConverter; a single custom converter I can register in the user data serializer; or zero or more converters to register in the user data serializer?

AArnott:
> I'm considering an alternative: making two MessagePackSerializer instances, one for the message envelopes, and one for user data

That makes a lot of sense. It's the closest analogy to the distinct `MessagePackSerializationOptions` it did for MessagePack-CSharp. 
And just like for that case, you'd still need to inject a few custom converters into the user data instance to handle the exotic types.

> Would it make more sense for the user to provide a shape provider (my current thinking) that gets passed to call to Serialize and GetConverter; a single custom converter I can register in the user data serializer; or zero or more converters to register in the user data serializer?

The user _should_ be able to register their own converters (because that provides more customizeability than just providing a type shape). And I suppose there is likely a need for them to provide their own `ITypeShapeProvider` that we would then use when we make the direct serialization calls.

Issue #198, opened 2024-12-24
Add analyzer for writing MessagePackString


trippwill:
I missed a few cases where I wasn't calling writer.WriteRaw(MyMPString.MsgPack.Span) correctly. An analyzer that would flag these could be useful and save time debugging.

AArnott:
When you say you weren't calling it correctly, what were you doing instead (that the analyzer would look for)?

trippwill:
There are many call patterns that continue to work but may not be the best way, such as Write(MyMPString.Value), or the more experimental Write(MyMPString.Utf8.Span), WriteRaw(MyMPString.Utf8.Span).

I actually haven't got the round trip working yet using TryRead.

One suggestion for the writer API, maybe Write() could take a MessagePackString directly?

PR #199, opened 2024-12-24
Add friendlier `MessagePackWriter.Write(MessagePackString)` method

AArnott:
Hopefully reduce the need for the analyzer requested in #198.


AArnott:
FYI @trippwill

trippwill:
I think this is a good alternative to #198

trippwill:
I think this is a good alternative to #198

PR #209, opened 2024-12-29
Gracefully handle ctor parameters named by the serialized name

### LINKED ISSUE #208: Duplicate Key Exception when serializing struct with PropertyShape attributes
Version: CI 0.3.76-beta

I haven't found time to reduce the repro, but I thought I should go ahead and report what I saw.

For this struct:
```csharp
    [DataContract]
    internal struct MarshalToken
    {
        [MessagePack.SerializationConstructor]
#pragma warning disable SA1313 // Parameter names should begin with lower-case letter
        public MarshalToken(int __jsonrpc_marshaled, long handle, string? lifetime, int[]? optionalInterfaces)
#pragma warning restore SA1313 // Parameter names should begin with lower-case letter
        {
            this.Marshaled = __jsonrpc_marshaled;
            this.Handle = handle;
            this.Lifetime = lifetime;
            this.OptionalInterfacesCodes = optionalInterfaces;
        }

        [DataMember(Name = "__jsonrpc_marshaled", IsRequired = true)]
        [STJ.JsonPropertyName("__jsonrpc_marshaled"), STJ.JsonRequired]
        public int Marshaled { get; set; }

        [DataMember(Name = "handle", IsRequired = true)]
        [STJ.JsonPropertyName("handle"), STJ.JsonRequired]
        public long Handle { get; set; }

        [DataMember(Name = "lifetime", EmitDefaultValue = false)]
        [STJ.JsonPropertyName("lifetime"), STJ.JsonIgnore(Condition = STJ.JsonIgnoreCondition.WhenWritingNull)]
        public string? Lifetime { get; set; }

        [DataMember(Name = "optionalInterfaces", EmitDefaultValue = false)]
        [STJ.JsonPropertyName("optionalInterfaces"), STJ.JsonIgnore(Condition = STJ.JsonIgnoreCondition.WhenWritingNull)]
        public int[]? OptionalInterfacesCodes { get; set; }
    }
```

I tried applying the PropertyShape attribute to each property, using the same names as the other serialization attributes. Theses name match the ctor argument names.

I also applied the GenerateShape attribute, and made the struct partial.

When serializing using NBMP, I would get a duplicate key exception. I traced the source to NB.MP GetOrCreateConverter, on a line that produced a dictionary from the type shape's constructor params. 

The solution was to remove the PolyType attributes from the struct, and use a witness type for the struct's shape.

AArnott:
Fixes #208

### DIFF
--- a/src/Nerdbank.MessagePack/StandardVisitor.cs
+++ b/src/Nerdbank.MessagePack/StandardVisitor.cs
@@ -83,7 +83,7 @@ internal class StandardVisitor : TypeShapeVisitor, ITypeShapeFunc
 		Dictionary<string, IConstructorParameterShape>? ctorParametersByName = null;
 		if (ctorShape is not null)
 		{
-			ctorParametersByName = new(StringComparer.OrdinalIgnoreCase);
+			ctorParametersByName = new(StringComparer.Ordinal);
 			foreach (IConstructorParameterShape ctorParameter in ctorShape.Parameters)
 			{
 				// Keep the one with the Kind that we prefer.
@@ -331,7 +331,7 @@ internal class StandardVisitor : TypeShapeVisitor, ITypeShapeFunc
 						return new ObjectArrayConverter<TDeclaringType>(inputs.GetJustAccessors(), constructorShape.GetDefaultConstructor(), !this.owner.SerializeDefaultValues);
 					}
 
-					Dictionary<string, int> propertyIndexesByName = new(StringComparer.OrdinalIgnoreCase);
+					Dictionary<string, int> propertyIndexesByName = new(StringComparer.Ordinal);
 					for (int i = 0; i < inputs.Properties.Count; i++)
 					{
 						if (inputs.Properties[i] is { } property)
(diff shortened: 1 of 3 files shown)


src/Nerdbank.MessagePack/StandardVisitor.cs
```diff
@@ -80,8 +80,23 @@ internal StandardVisitor(ConverterCache owner, TypeGenerationContext context)
 
 		IConstructorShape? ctorShape = objectShape.Constructor;
 
-		Dictionary<string, IConstructorParameterShape>? ctorParametersByName =
-			ctorShape?.Parameters.ToDictionary(p => p.Name, StringComparer.OrdinalIgnoreCase);
+		Dictionary<string, IConstructorParameterShape>? ctorParametersByName = null;
+		if (ctorShape is not null)
+		{
+			ctorParametersByName = new(StringComparer.OrdinalIgnoreCase);
```
eiriktsarpalis: Is hardcoding case insensitive matching desirable in the general case?

AArnott: Good question. I tried changing this to `Ordinal` (and one other place) and all my tests still pass. So I guess I'll switch to `Ordinal` since that seems safer, until I have tests that prove that's undesirable. Thanks for the suggestion.

PR #221, opened 2025-01-16
Add test for class with indexer

AArnott:


eiriktsarpalis:
Was this missing from the PolyType test cases?

Issue #278, opened 2025-02-18
Overcome nested subtype aliasing limitation


AArnott:
From [Unions](https://aarnott.github.io/Nerdbank.MessagePack/docs/unions.html?tabs=net), I have this sentence in the document:

> It will not do for `Animal` to merely mention `Horse` and for `Horse` to list `Quarterback` as a sub-type, as this is not currently supported. 

A technical challenge prevented me from achieving nested encoding, but it is something I'd like to achieve since it makes intuitive sense.

Given a Quarterhorse with alias 1 defined on its base Horse class, and Horse with alias X on its base Animal class, someone serializing an Animal that happened to a Quarterhorse would (ideally) get: 

```json
["X",[1,"quarterhorse-property":"something"]]
```

Note how X navigates to Horse and 1 navigates further. This totally makes sense to me... I just couldn't implement it at the time due to how I did known subtype conversion.

Issue #267, opened 2025-02-10
Customizing serialization - force array format


joelverhagen:
Is there a way to force `MessagePackSerializer` to always serialize objects using the array format?

From this doc:
https://aarnott.github.io/Nerdbank.MessagePack/docs/customizing-serialization.html#array-or-map

> When [MessagePackSerializer.SerializeDefaultValues](https://aarnott.github.io/Nerdbank.MessagePack/api/Nerdbank.MessagePack.MessagePackSerializer.html#Nerdbank_MessagePack_MessagePackSerializer_SerializeDefaultValues) is set to [SerializeDefaultValuesPolicy.Always](https://aarnott.github.io/Nerdbank.MessagePack/api/Nerdbank.MessagePack.SerializeDefaultValuesPolicy.html#Nerdbank_MessagePack_SerializeDefaultValuesPolicy_Always), the array format is always chosen, even if gaps from unassigned indexes may exist in the array.

It seems like this would be the way, but perhaps this is dependent on `[Key(...)]` attributes to be applied. It doesn't work the way I guessed:

```csharp
using Nerdbank.MessagePack;
using PolyType;

// serialize data
var stream = new MemoryStream();

var nerdbankSerializer = new MessagePackSerializer { SerializeDefaultValues = SerializeDefaultValuesPolicy.Always };
var record = new PersonA("Bill Gates", 42);
nerdbankSerializer.Serialize(stream, record);

// deserialize data
stream.Position = 0;
Console.WriteLine(MessagePackSerializer.ConvertToJson(stream.ToArray()));

[GenerateShape]
public partial record PersonA(string Name, int Age);
```

Output:
```
{"Name":"Bill Gates","Age":42}
```

Ideally there is a way to force array format without `[Key(...)]` because then I would need to manually set `KeyAttribute` on all of my properties, very much a kind of mechanical work I'd like to avoid and can't easily generate due to source generators not being able to depend on each other.

[...]

PR #290, opened 2025-03-02
Adopt PolyType's own union type support

AArnott:
PolyType 0.28.1 makes two major changes, which this commit adapts to:

* Renames Nullable type shapes to Optional.
* Provides a new `DerivedTypeShapeAttribute`, which removes the need for our own `KnownSubTypeAttribute` and analyzers.

Fixes https://github.com/AArnott/Nerdbank.MessagePack/issues/278


AArnott:
This is blocked by PolyType providing a way to tell whether `Tag` is explicitly specified on the new `DerivedTypeShapeAttribute`, which isn't directly what https://github.com/eiriktsarpalis/PolyType/issues/121 is all about but a related discussion is happening there.
CC: @eiriktsarpalis

src/Nerdbank.MessagePack/StandardVisitor.cs
```diff
@@ -174,7 +172,65 @@ internal StandardVisitor(ConverterCache owner, TypeGenerationContext context)
...
+			Getter<TUnion, int> getUnionCaseIndex = unionShape.GetGetUnionCaseIndex();
+			Dictionary<int, MessagePackConverter> deserializerByIntAlias = new(unionShape.UnionCases.Count);
+			List<(DerivedTypeIdentifier Alias, MessagePackConverter Converter, ITypeShape Shape)> serializers = new(unionShape.UnionCases.Count);
+			KeyValuePair<int, MessagePackConverter<TUnion>>[] unionCases = unionShape.UnionCases
+				.Select(unionCase =>
+				{
+					// TODO: Use the Name if Tag isn't set explicitly for better schema stability.
+					// Poor man's substitute: if Tag != Index, we know the user set it explicitly.
+					// But they might have set it explicitly to be the same as Index, so we need a feature from PolyType.
+					DerivedTypeIdentifier alias = unionCase.Tag != unionCase.Index ? new(unionCase.Tag) : new(unionCase.Name);
```
AArnott: This needs an `IUnionCaseShape.IsTagSpecified` property or similar so that we can use the tag _iff_ the user explicitly specified one. Using an implicitly specified one is no good unless the user has opted into an "unstable schema" mode (which we may introduce for https://github.com/aarnott/nerdbank.MessagePack/issues/267).

eiriktsarpalis: If you want to play it safe, you could also just resolve the attribute at runtime and check if the value equals -1:

https://github.com/eiriktsarpalis/PolyType/blob/main/src/PolyType/DerivedTypeShapeAttribute.cs#L44

AArnott: That's reasonable. It requires I search all the attributes for a match, but it works.
https://github.com/AArnott/Nerdbank.MessagePack/compare/74bc99f76125f3954b3e5f3e4e93e87c8516b9e1..e93dc61de1b9ecb8b96719e3ef6db2977d6161fa

PR #298, opened 2025-03-06
Simplify F# sample

eiriktsarpalis:


samples/fs/FSharpSamples.fsproj
```diff
@@ -4,12 +4,17 @@
...
     <Nullable>disable</Nullable>
+    <OutputType>Exe</OutputType>
   </PropertyGroup>
 
   <ItemGroup>
     <Compile Include="Program.fs" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="FSharp.Core" />
```
eiriktsarpalis: Not sure why, but running the project is failing unless I add this reference. Doesn't typically happen since this dependency is filled by the sdk.

AArnott: It's not required on _my_ machine, so I'm going to drop this from your change before merging and let you sort it out. :)

My original sample had been an exe. I changed it to a library to match the C# samples, which was useful because I have a great many C# samples so having any one of them execute doesn't make sense. For F#, I anticipate having more samples. But for now at least, sure, let's make it an exe.

AArnott (approved):
Thanks.

PR #310, opened 2025-03-13
Get open generic custom converters working

### LINKED ISSUE #182: Support for custom converters of open generic data types
Support I have this data type:

```cs
public partial record GenericData<T>
{
   public T? Value { get; set; }
}
```

This type is not allowed to have `[GenerateShape]` applied to it, I notice. Fair enough. It'll be referenced as a closed generic type eventually somewhere else and get a shape that way.  And if it can be serialized automatically, that's fantastic.

But suppose I then declare a custom converter for it:

```cs
class GenericDataConverter<T> : MessagePackConverter<GenericData<T>>
```

Now I have a problem, because AOT support forbids me from calling `Type.MakeGenericType`, but how else am I going to get an instance of the converter that I need?

There are a few scenarios I have in mind:

1. The `MessagePackConverterAttribute`, which may be applied this way:
   ```cs
   [MessagePackConverter(typeof(GenericDataConverter<>))]
   public partial record GenericData<T>
   {
      public T? Value { get; set; }
   }
   ```
2. Runtime converter registration of closed generic types:
   ```cs
   MessagePackSerializer.RegisterConverter(new GenericDataConverter<string>());
   ```
3. Runtime converter registration of open generic types:
   ```cs
   MessagePackSerializer.RegisterConverter(typeof(GenericDataConverter<>));
   ```
   or perhaps:
   ```cs
   MessagePackSerializer.RegisterConverter(typeof(GenericData<>));
   ```

Right now we have no support for open generic converters at all, but that's blocking using this library from vs-StreamJsonRpc which supports serializing things like `IProgress<T>` (for arbitrary, though presumably shaped, T's).
We _do_ potentially have a statically discoverable source for the closed `IProgress<T>` type, but there is currently no source generator that knows to generate code to construct a closed `ProgressConverter<T>` type for each `T` it finds statically.

AArnott:
Closes #182

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -7,7 +7,7 @@
     <RoslynVersion>4.13.0</RoslynVersion>
     <RoslynVersionForAnalyzers>4.11.0</RoslynVersionForAnalyzers>
     <CodeAnalysisAnalyzerVersion>3.11.0-beta1.24629.2</CodeAnalysisAnalyzerVersion>
-    <PolyTypeVersion>0.29.3</PolyTypeVersion>
+    <PolyTypeVersion>0.29.3-g4b0aeb5295</PolyTypeVersion>
   </PropertyGroup>
   <ItemGroup>
     <PackageVersion Include="BenchmarkDotNet" Version="0.14.0" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -6,8 +6,8 @@
     <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
     <RoslynVersion>4.13.0</RoslynVersion>
     <RoslynVersionForAnalyzers>4.11.0</RoslynVersionForAnalyzers>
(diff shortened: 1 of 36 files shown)


AArnott:
This is currently blocked on the merge of https://github.com/eiriktsarpalis/PolyType/pull/128

Directory.Packages.props
```diff
@@ -7,7 +7,7 @@
     <RoslynVersion>4.13.0</RoslynVersion>
     <RoslynVersionForAnalyzers>4.11.0</RoslynVersionForAnalyzers>
     <CodeAnalysisAnalyzerVersion>3.11.0-beta1.24629.2</CodeAnalysisAnalyzerVersion>
-    <PolyTypeVersion>0.29.3</PolyTypeVersion>
+    <PolyTypeVersion>0.29.3-g4b0aeb5295</PolyTypeVersion>
```
AArnott: This is currently blocked on the merge of https://github.com/eiriktsarpalis/PolyType/pull/128

PR #332, opened 2025-03-22
Add `DeserializeDynamic` method for typeless deserialization

### LINKED ISSUE #330: Add "primitives" deserialization API
MessagePack is made up of primitives, maps, vectors and extensions. All but the extensions can be trivially deserialized anyway into equivalent .NET primitives (with lots of boxing). 

In MessagePack-CSharp it looks like this:

```cs
MessagePackSerializer.Deserialize<object>(this.MsgPackData, this.serializerOptions.WithResolver(PrimitiveObjectResolver.Instance));
```

StreamJsonRpc leverages this when deserializing error details. We may need to expose a similar API.

AArnott:
Adds the following method to `MessagePackSerializer` for typeless deserialization:

```cs
public static dynamic? DeserializeDynamic(ref MessagePackReader reader)
```

This allows for javascript-like decoding:

```cs
dynamic deserialized = MessagePackSerializer.DeserializeDynamic(ref reader)!;
Assert.NotNull(deserialized);
Assert.Equal("Value1", deserialized.Prop1);
Assert.Equal(42, (int)deserialized.Prop2);
Assert.Equal(3, deserialized.deeper.Length);
Assert.Equal(true, deserialized.deeper[0]);
Assert.Equal(3.5, deserialized.deeper[1]);
Assert.Equal(new Extension(15, new byte[] { 1, 2, 3 }), deserialized.deeper[2]);
Assert.Equal(new byte[] { 1, 2, 3 }, deserialized[45UL]);
```

Closes #330

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -17,6 +17,7 @@
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="$(RoslynVersion)" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.CodeFix.Testing.XUnit" Version="1.1.2" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="$(RoslynVersion)" />
+    <PackageVersion Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageVersion Include="Microsoft.NET.StringTools" Version="17.13.9" />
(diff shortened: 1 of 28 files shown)


AArnott:
Considering [MessagePack-CSharp's support for `ExpandoObject`](https://github.com/MessagePack-CSharp/MessagePack-CSharp/blob/master/doc/ExpandoObject.md), we should revise the docs change in this PR and consider exposing this functionality in a similar way that doesn't require a dedicated method. And maybe support _serialization_ as well.

PR #337, opened 2025-03-24
Add `MessagePackValue` variant struct and converter

AArnott:
I don't know yet whether this will be useful.


AArnott:
@eiriktsarpalis you suggested I craft a DOM type. This is my attempt. Any reactions?

AArnott:
So weird... my new struct fails to load as a type on .NET Framework. I wonder why. I'm not doing anything fancy.

AArnott:
The .NET Framework test failure has been addressed.

PR #356, opened 2025-04-16
Add support for backward/forward version compatibility data preservation

AArnott:
Adds an `IVersionSafeObject` interface which user-defined data types may implement in order to retain data from a deserialized format that they don't recognize, so that it may be retained rather than lost if that object is later serialized.

- [x] Measure perf impact
- [x] Update doc with 'sparse' declaration relay use case
- [x] ~~Consider exposing a lazily-decoded MessagePackValue tree for each retained unused value.~~
- [x] Add test for a struct that implements the new interface.
- [x] Switch from interface to ordinary property, which must be part of the 'shape', so that structs work.


AArnott:
@eiriktsarpalis any feedback on the names for the added public API?

PR #417, opened 2025-05-30
Add trim feature switch to shrink output size

### LINKED ISSUE #406: Review binary size of AotNativeConsole test/sample
The `AotNativeConsole` sample is 37MB when published for NativeAOT win-x64. That seems excessive. Let's review whether this regressed over the last few months and what we can do in the library to reduce this substantially.

AArnott:
#335 regressed our AotNativeConsole.exe sample from 4.9MB output to 11.99MB.
All it did was add 3 converters for System.Text.Json types.

This change allows applications that do not require these converters to avoid the trimmed output size penalty of the library having this feature by turning it off.

Fixes #406.


AArnott:
This seems a simple matter of leveraging .NET 9's feature switches to remove these 3 converters when the application does not require them. But for some reason, it is entirely ineffective. The output size of `dotnet publish` is still 11.4MB. 

@eiriktsarpalis do you have any idea why?

eiriktsarpalis:
@sbomer might be able to help here, I see they introduced the [PR](https://github.com/dotnet/runtime/pull/106564) replacing ILLink.Substitutions declarations with `FeatureSwithDefinition` attributes.

test/AotNativeConsole/AotNativeConsole.csproj
```diff
@@ -13,6 +13,10 @@
     <IlcGenerateDgmlFile>true</IlcGenerateDgmlFile>
   </PropertyGroup>
 
+  <ItemGroup>
+    <RuntimeHostConfigurationOption Include="Feature.MessagePack.SystemTextJsonConverters" Value="false" />
```
sbomer:
```suggestion
    <RuntimeHostConfigurationOption Include="Feature.MessagePack.SystemTextJsonConverters" Value="false" Trim="true" />
```

For trimming support we require you to set Trim="true" on this. I see this is missing from the docs at https://learn.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.featureswitchdefinitionattribute, will get that fixed!

AArnott: Yup, that did it. Thanks.

PR #418, opened 2025-05-30
Fix serializing large values to a Stream

AArnott:


Arma-Branca:
> @Arma-Branca this doesn't repro the issue you reported. Can you help beef it up, or provide a repro that demonstrates the problem?

I'm using the following serializer without using witnesses in that test

`
new MessagePackSerializer { SerializeDefaultValues = SerializeDefaultValuesPolicy.Never, DeserializeDefaultValues = DeserializeDefaultValuesPolicy.AllowMissingValuesForRequiredProperties }, ReflectionTypeShapeProvider.Default)
`

AArnott:
@Arma-Branca Thanks. I have a repro now and will work on a fix.

PR #453, opened 2025-06-09
Add serde.NET to the feature comparison table

AArnott:


AArnott:
@agocke would you care to review this? I want to report accurately.

agocke:
Happy to, thanks. Do you have a section on unsafe code? That might be relevant from a security perspective.

AArnott:
> Do you have a section on unsafe code?

No. Would it cover what the library uses internally, or what it source generates?

And what is unsafe code? C#'s `unsafe` keyword comes to mind, but I've seen very unsafe code written without `unsafe` and plenty of good, safe code within `unsafe` blocks. I don't think I'd want to contribute to the misconception that everything is safe unless you use `unsafe`. But I suppose within source generated code, knowing whether the `unsafe` keyword is used is important because it means the csproj will need to be changed to allow it.

AArnott:
@agocke for your convenience in reviewing, here is the [rendered version](https://github.com/AArnott/Nerdbank.MessagePack/blob/serde-dotnet-comparison/docfx/docs/features.md#feature-comparison).

agocke:
Re unsafe, I think this is relevant context: https://github.com/dotnet/designs/pull/330. In other words, yes there are lots of things that may play into unsafe, even though they're not currently marked as such

docfx/docs/features.md
```diff
@@ -23,38 +23,38 @@ See how this library compares to other .NET MessagePack libraries.
...
+Skip serializing default values | [✅](xref:Nerdbank.MessagePack.MessagePackSerializer.SerializeDefaultValues) | [❌](https://github.com/MessagePack-CSharp/MessagePack-CSharp/issues/678) | 🌗 |
+Required and non-nullable property deserialization guaranteed | [✅](xref:Nerdbank.MessagePack.MessagePackSerializer.DeserializeDefaultValues) | ❌ | ✅ |
+Dynamically use maps or arrays for most compact format | [✅](customizing-serialization.md#array-or-map) | [❌](https://github.com/MessagePack-CSharp/MessagePack-CSharp/issues/1953) | ❌ |
+Surrogate types for automatic serialization of unserializable types | [✅](surrogate-types.md) | ❌ | [✅](https://serdedotnet.github.io/foreign-types.html) |
+Custom converters         | [✅](custom-converters.md) | ✅ | [✅](https://serdedotnet.github.io/customization.html)
+Stateful converters       | [✅](custom-converters.md#stateful-converters) | ❌ | ❌ |
+Deserialization callback  | [✅](xref:Nerdbank.MessagePack.IMessagePackSerializationCallbacks) | [✅](https://github.com/MessagePack-CSharp/MessagePack-CSharp?tab=readme-ov-file#serialization-callback) | ❌ |
```
agocke, line 41: Hmm, how does this work? How can you call an instance method on an object before it has been deserialized?

docfx/docs/features.md
```diff
@@ -23,38 +23,38 @@ See how this library compares to other .NET MessagePack libraries.
...
+MsgPack extensions        | ✅ | ✅ | ❌ |
+LZ4 compression           | [❌](https://github.com/AArnott/Nerdbank.MessagePack/issues/34) | [✅](https://github.com/MessagePack-CSharp/MessagePack-CSharp?tab=readme-ov-file#lz4-compression) | ❌ |
+Trim-safe                 | ✅ | ❌ | ✅ |
+NativeAOT ready           | ✅ | ❌[^2] | ✅ |
+Unity                     | [✅](unity.md)[^3] | [✅](https://github.com/MessagePack-CSharp/MessagePack-CSharp?tab=readme-ov-file#unity-support) | ❔ |
+Async                     | [✅](xref:Nerdbank.MessagePack.MessagePackSerializer.SerializeAsync*) | ❌ | ❌ |
```
agocke, line 47: This is correct, although I plan to add this eventually.

docfx/docs/features.md
```diff
@@ -23,38 +23,38 @@ See how this library compares to other .NET MessagePack libraries.
...
+Endless streaming deserialization | [✅](streaming-deserialization.md) | [✅](https://github.com/MessagePack-CSharp/MessagePack-CSharp/?tab=readme-ov-file#multiple-messagepack-structures-on-a-single-stream) | ❌ |
+Reference preservation    | [✅](xref:Nerdbank.MessagePack.MessagePackSerializer.PreserveReferences) | ❌ | ❌ |
+Cyclical references       | [✅](xref:Nerdbank.MessagePack.MessagePackSerializer.PreserveReferences) | ❌ | ❌ |
+JSON schema export        | [✅](xref:Nerdbank.MessagePack.MessagePackSerializer.GetJsonSchema*) | ❌ | ❌ |
+Secure defaults           | [✅](security.md) | ❌ | ❌ |
```
agocke, line 52: Is an example of a problematic structure `Dictionary<int, int>` because it's not collision-resistant?

docfx/docs/features.md
```diff
@@ -23,38 +23,38 @@ See how this library compares to other .NET MessagePack libraries.
...
+Automatic hash collection deserialization in secure mode | ✅ | ✅ | ❌ |
+Automatic collision-resistant hash function for arbitrary types | [✅](xref:Nerdbank.MessagePack.StructuralEqualityComparer) | ❌ | ❌ |
+Rejection of data that defines multiple values for the same property | [✅](security.md#multiple-values-for-the-same-property) | ❌ | ❌ |
```
agocke, line 55: I also plan to fix this.

agocke (approved):
Looks right to me, for now. Some stuff may change assuming I get more time to work on this

docfx/docs/features.md
AArnott: We invoke the callbacks between running the constructor and setting the properties.
This has its limitations, of course. And in particular with PolyType, once a type has `required` or `init` properties, all property sets are grouped within the pseudo-constructor, so we don't do an OnBeforeDeserialization callback there. But OnAfterDeserialization of course works.

docfx/docs/features.md
AArnott: That's not an example I would use because I don't know how well `int` as a key works in the dictionary for collision resistance. But that might be a good one. 
I would tend to all the other types, including user-defined types, for which GetHashCode is implemented without collision resistance in mind.

AArnott:
Thanks for reviewing. I'll be happy to update (or take PRs) to update this column as you make updates to serde.NET.

PR #465, opened 2025-06-24
Throw on serializing `DateTimeKind.Unspecified`

AArnott:
MessagePack-CSharp suffers from accidental datetime corruption for `Unspecified` by assuming it is UTC. If it was actually local time, it is corrupted. To avoid this being an issue for the new Nerdbank.MessagePack library, just throw when that ambiguity is encountered, forcing users to specify which Kind it is.


AArnott:
Thanks to @AlaskanDruid for calling out this unintuitive behavior in time for it to be solved in the new library.

PR #506, opened 2025-07-23
Add `UseComparerAttribute`

### LINKED ISSUE #505: Offer comparer-provider attribute for collection properties
We should offer an attribute that can conveniently indicate that a collection should be deserialized with a particular key comparer.

For example:

```cs
class Foo
{
   [UseComparer(typeof(StringComparer), nameof(StringComparer.OrdinalIgnorecase))]
   public Dictionary<string, Something> MyCaseInsensitiveDictionary { get; set; }
}
```

This attribute should be applicable to properties, fields and parameters.
It should have these constructors available:

```cs
   public UseComparerAttribute(Type comparerType);
   public UseComparerAttribute(Type comparerType, string memberName);
```

A diagnostic analyzer should be created that verifies that the Type or Type/Member pair specified in the attribute point to an `IEqualityComparer<T>` or `IComparer<T>` implementation suitable for the attributed collection type.

At runtime we look for this attribute and customize the `CollectionConstructionOptions<TKey>` struct to provide that comparer when constructing the collection. 
This might mean a change to the `StandardVisitor.GetCollectionOptions` method.

The tricky bit is that right now, attributes on a member don't change how we deserialize the type itself at all. We'll have to find some way to plumb that information through.

AArnott:
Closes #505

### DIFF
--- a/docfx/analyzers/NBMsgPack070.md
+++ b/docfx/analyzers/NBMsgPack070.md
@@ -1,0 +1,13 @@
+# NBMsgPack070: UseComparerAttribute type must not be an open generic
+
+<xref:Nerdbank.MessagePack.UseComparerAttribute> should specify a concrete type, not an open generic type.
+
+## Example violation
+
+[!code-csharp[](../../samples/cs/AnalyzerDocs/NBMsgPack070.cs#Defective)]
+
+## Resolution
+
+Specify a concrete type instead of an open generic type:
+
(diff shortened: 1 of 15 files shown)


AArnott:
@eiriktsarpalis What do you think? This is a lot of code to support something that can _almost_ always be done more simply just by removing the `set` accessor and adding an initializer that uses the intended comparer. The only thing this really unblocks is literal constructor parameters.

PR #509, opened 2025-07-26
Bump PolyType to 0.50.1 and switch to its new parameter tracking

AArnott:
This also removes code that was buggy in tracking required parameters for objects with more than 64 parameters, as can be seen by the updated test (which had been verifying buggy behavior).


AArnott:
CC: @eiriktsarpalis

PR #576, opened 2025-08-28
Document limitations of nullability enforcement

AArnott:


src/Nerdbank.MessagePack/DeserializeDefaultValuesPolicy.cs
```diff
@@ -13,6 +13,13 @@ public enum DeserializeDefaultValuesPolicy
 	/// Deserialization will fail if any required properties have no specified values in the data stream,
 	/// or when <see langword="null" /> values are encountered for non-nullable properties.
 	/// </summary>
+	/// <remarks>
+	/// The non-nullability enforcement by the deserializer is limited to non-generic parameters, properties and fields.
```
AArnott: @eiriktsarpalis What is it about generic typed fields, properties and parameters that breaks non-nullability? For example, isn't a property typed as `List<string>` as easy to note as nullable or non-nullable as one typed as `MyObject` would be?

eiriktsarpalis: A property typed as `List<string>` shouldn't have any issues surfacing the right annotation. The issue is when you have a generic property as follows:
```C#
public class Foo<T>
{
    T Value { get; set; }
}
```
The nullability of `Foo<string>.Value` would never be surfaced as non-nullable because at run time it is indistinguishable from `Foo<string?>.Value`.

AArnott: Ah, ok. So I think I'll rephase as fields, properties and parameters typed as generic type parameters then.

PR #584, opened 2025-08-31
Use custom msgpack extensions for .NET "primitive" types

### LINKED ISSUE #574: "Native" binary encoding of decimal (and other .Net types)
Do you have any plans to add compact binary converters for primitive .net types not natively supported by messagepack similar to 
https://github.com/MessagePack-CSharp/MessagePack-CSharp?tab=readme-ov-file#use-native-resolvers ?

It is not to difficult to do ourselves, but it would be nice to know if this fits your vision so we can choose an good encoding (BIN8 or FixExt16).
We currently only need decimal, but I think it could make sense for (U)int128 as well.

If so what design would you like for the features public API? 

Separate extension points (such as for guid) for each type such as ;
```csharp
WithDecimalConverter(this MessagePackSerializer serializer, OptionalConverters.DecimalFormat format = DecimalFormat.BinaryLittleEndian)
```

What kind of binary format (BIN8 or new FixExt16 +TypeCodes) ? 

If you think this makes sense @fremo917 can se if we get the time to submit a PR.

We are currently using something similar to the below code (but needs to be changed if it is to support big endian platforms)

```csharp

	/// <summary>
	/// Formatter which writes a struct as raw bytes instead of strings.
	/// Useful for Decimal and guid that have fixed size
	/// </summary>
	internal class RawStructConverter<T> : MessagePackConverter<T>
		where T : unmanaged
	{
		public RawStructConverter()
		{
			Debug.Assert(BitConverter.IsLittleEndian);
		}

		public unsafe override T Read(ref MessagePackReader reader, SerializationContext context)
		{
			var bytes = reader.ReadBytes();

			if (bytes is not null && bytes.Value.Length == sizeof(T))
			{
				T result;
				var dest = new Span<byte>(&result, sizeof(T));
				bytes.Value.CopyTo(dest);
				return result;
			}

			throw new MessagePackSerializationException("invalid data for decimal");
		}

		public override void Write(ref MessagePackWriter writer, in T value, SerializationContext context)
		{
			var rawBytes = MemoryMarshal.Cast<T, byte>(new ReadOnlySpan<T>(in value));
			writer.Write(rawBytes);
			// CONSIDER: using custom extension with typecode in range ( 64, 127)
			// - will be encoded as (FixExt16, typecode, data) if size is 16 bytes
			// writer.Write(new Extension(65, rawBytes));
			// - The current approach encodes it as (Bin8, length: (byte)3, data)
		}
	}

	/// <summary>
	/// Formatter which writes Decimal as raw bytes instead of strings, base on MessagePack-CSharp
	/// https://github.com/MessagePack-CSharp/MessagePack-CSharp/blob/master/src/MessagePack/Formatters/UnsafeBinaryFormatters.cs#L65
	/// </summary>
	internal sealed class RawDecimalConverter : RawStructConverter<decimal>
	{
	}
```

### LINKED ISSUE #579: Review .NET type converters for Bin vs. Ext encoding
We should revisit all our built-in converters for use of the String and Bin encoding and reconsider whether Ext would be a more appropriate encoding.

- BigInteger (binary)
- Decimal (string)
- Guid (string or binary)
- Int128 / UInt128

AArnott:
- Future proof the design of library extension type codes
- Encode `Int128` and `UInt128` as Bnative integers if they fit, otherwise as a BE msgpack extension. Allow decoding a LE Bin for compatibility with MessagePack-CSharp.
- Write `BigInteger` as an ordinary integer if it fits, otherwise use an extension. Allow decoding a Bin for compatibility with MessagePack-CSharp.
- Encode `decimal` as a msgpack extension instead of as a string. Allow decoding a string for compatibility with MessagePack-CSharp.
- Switch binary Guid encoding over to a msgpack extension
- Document the encodings in the MessagePack-CSharp migration section.

Closes #579
Closes #574

### DIFF
--- a/src/Nerdbank.MessagePack/Converters/PrimitiveConverters.cs
+++ b/src/Nerdbank.MessagePack/Converters/PrimitiveConverters.cs
@@ -947,7 +947,8 @@ internal class GuidAsLittleEndianBinaryConverter : MessagePackConverter<Guid>
 	public override Guid Read(ref MessagePackReader reader, SerializationContext context)
 	{
 		sbyte typeCode = LibraryReservedMessagePackExtensionTypeCode.ToByte(context.ExtensionTypeCodes.GuidLittleEndian);
-		ReadOnlySequence<byte> bytes = reader.ReadExtension(typeCode);
+		ReadOnlySequence<byte> bytes = reader.NextMessagePackType == MessagePackType.Binary
+			? reader.ReadBytes()!.Value : reader.ReadExtension(typeCode);
 
 		if (bytes.IsSingleSegment)
 		{
--- a/src/Nerdbank.MessagePack/Converters/PrimitiveConverters.cs
+++ b/src/Nerdbank.MessagePack/Converters/PrimitiveConverters.cs
@@ -599,7 +599,8 @@ internal class BigIntegerConverter : MessagePackConverter<BigInteger>
 			return MessagePackCode.IsSignedInteger(reader.NextCode) ? (BigInteger)reader.ReadInt64() : (BigInteger)reader.ReadUInt64();
 		}
 
-		ReadOnlySequence<byte> bytes = reader.ReadExtension(typeCode);
+		ReadOnlySequence<byte> bytes = reader.NextMessagePackType == MessagePackType.Binary
+			? reader.ReadBytes()!.Value : reader.ReadExtension(typeCode);
 #if NET
 		if (bytes.IsSingleSegment)
 		{
--- a/src/Nerdbank.MessagePack/Converters/PrimitiveConverters.cs
+++ b/src/Nerdbank.MessagePack/Converters/PrimitiveConverters.cs
@@ -315,6 +315,7 @@ internal class DecimalConverter : MessagePackConverter<decimal>
 	private const int MaxDecimalStringLength = 31; // -79228162514264337593543950335. // any decimals would trade sigfigs for length
 
 	/// <inheritdoc/>
+#pragma warning disable NBMsgPack031 // only read one structure
 	public override decimal Read(ref MessagePackReader reader, SerializationContext context)
 #pragma warning restore NBMsgPack031 // only read one structure
 	{
@@ -448,16 +449,23 @@ internal class Int128Converter : MessagePackConverter<Int128>
 	/// </summary>
 	internal static readonly Int128Converter Instance = new();
 
+	private const int Size = 128 / 8;
+
 	/// <inheritdoc/>
+#pragma warning disable NBMsgPack031 // only read one structure
 	public override Int128 Read(ref MessagePackReader reader, SerializationContext context)
+#pragma warning restore NBMsgPack031 // only read one structure
 	{
 		// Fail fast if the user hasn't reserved a type code these values,
 		// even if this particular value might fit in a native integer type.
 		sbyte typeCode = LibraryReservedMessagePackExtensionTypeCode.ToByte(context.ExtensionTypeCodes.Int128);
 
-		if (reader.NextMessagePackType == MessagePackType.Integer)
+		switch (reader.NextMessagePackType)
 		{
-			return MessagePackCode.IsSignedInteger(reader.NextCode) ? (Int128)reader.ReadInt64() : (Int128)reader.ReadUInt64();
+			case MessagePackType.Binary:
+				return this.ReadBin(ref reader, context);
+			case MessagePackType.Integer:
+				return MessagePackCode.IsSignedInteger(reader.NextCode) ? (Int128)reader.ReadInt64() : (Int128)reader.ReadUInt64();
 		}
 
 		ReadOnlySequence<byte> sequence = reader.ReadExtension(typeCode);
@@ -508,6 +516,33 @@ internal class Int128Converter : MessagePackConverter<Int128>
 			new JsonObject() { ["type"] = "integer" }),
 		["description"] = "A 128-bit signed integer",
 	};
+
+	[MethodImpl(MethodImplOptions.AggressiveInlining)]
+	private static bool TryReadLittleEndian<T>(ReadOnlySpan<byte> source, out T value)
+		where T : IBinaryInteger<T> => T.TryReadLittleEndian(source, isUnsigned: false, out value);
+
+	private Int128 ReadBin(ref MessagePackReader reader, SerializationContext context)
+	{
+		Debug.Assert(reader.NextMessagePackType == MessagePackType.Binary, "Our caller should have guaranteed this.");
+		ReadOnlySequence<byte> sequence = reader.ReadBytes()!.Value;
+		if (sequence.Length != Size)
+		{
+			throw new MessagePackSerializationException("Invalid Int128 data.");
+		}
+
+		if (sequence.IsSingleSegment)
+		{
+			TryReadLittleEndian(sequence.FirstSpan, out Int128 value);
+			return value;
+		}
+		else
+		{
+			Span<byte> bytes = stackalloc byte[Size];
+			sequence.CopyTo(bytes);
+			TryReadLittleEndian(bytes, out Int128 value);
+			return value;
+		}
+	}
 }
 
 /// <summary>
@@ -520,16 +555,23 @@ internal class UInt128Converter : MessagePackConverter<UInt128>
 	/// </summary>
 	internal static readonly UInt128Converter Instance = new();
 
+	private const int Size = 128 / 8;
+
 	/// <inheritdoc/>
+#pragma warning disable NBMsgPack031 // only read one structure
 	public override UInt128 Read(ref MessagePackReader reader, SerializationContext context)
+#pragma warning restore NBMsgPack031 // only read one structure
 	{
 		// Fail fast if the user hasn't reserved a type code these values,
 		// even if this particular value might fit in a native integer type.
 		sbyte typeCode = LibraryReservedMessagePackExtensionTypeCode.ToByte(context.ExtensionTypeCodes.UInt128);
 
-		if (reader.NextMessagePackType == MessagePackType.Integer)
+		switch (reader.NextMessagePackType)
 		{
-			return (UInt128)reader.ReadUInt64();
+			case MessagePackType.Binary:
+				return this.ReadBin(ref reader, context);
+			case MessagePackType.Integer:
+				return (UInt128)reader.ReadUInt64();
 		}
 
 		ReadOnlySequence<byte> sequence = reader.ReadExtension(typeCode);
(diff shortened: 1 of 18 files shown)


test/Nerdbank.MessagePack.Tests/BuiltInConverterTests.cs
```diff
@@ -44,10 +126,85 @@ public partial class BuiltInConverterTests : MessagePackSerializerTestBase
...
+	}
 
+	/// <summary>
+	/// Verifies that we can read <see cref="decimal"/> values that use the Bin header, which is what MessagePack-CSharp's "native" formatter uses.
+	/// </summary>
 	[Fact]
-	public void BigInteger() => this.AssertRoundtrip(new HasBigInteger(1));
+	public void Decimal_FromBin()
+	{
+		Span<decimal> value = [1.2m];
```
Daniel-Svensson: nit: 
  The test works fine, but you might want to use a decimal where all 4 ints are different such as new `decimal(ulong.MaxValue) * -1000` used above


FYI: If you want raw msg data (for decimal(ulong.MaxValue) * -1000)  then it is 
```csharp
ReadOnlySpan<byte> byteArray = [0xC4, 0x10, 0x00, 0x00, 0x00, 0x80, 0xE7, 0x03, 0x00, 0x00, 0x18, 0xFC, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF];
```


Produced by below code:
```csharp
decimal a = new decimal(ulong.MaxValue) * -1000;

MessagePackSerializerOptions options = new MessagePackSerializerOptions(NativeDecimalResolver.Instance);
var bytes = MessagePack.MessagePackSerializer.Serialize(a, options);

string formattedCode = $"ReadOnlySpan<byte> byteArray = [ {string.Join(", ", bytes.Select(b => $"0x{b:X2}"))} ];";
formattedCode.Dump();
```

test/Nerdbank.MessagePack.Tests/BuiltInConverterTests.cs
```diff
@@ -44,10 +126,85 @@ public partial class BuiltInConverterTests : MessagePackSerializerTestBase
 #endif
 
 	[Fact]
-	public void Decimal() => this.AssertRoundtrip(new HasDecimal(1.2m));
+	public void Decimal()
+	{
+		this.AssertRoundtrip(new HasDecimal(1.2m));
+		this.AssertRoundtrip(new HasDecimal(new decimal(ulong.MaxValue) * 1000));
+		this.AssertRoundtrip(new HasDecimal(new decimal(ulong.MaxValue) * -1000));
+	}
```
Daniel-Svensson, line 142: God test coverage :+1: especially the last one where all the bits (4 ints) are different

If I had made the changes myself I would probably also have added a comparison against hard-coded bytes of the expected msg pack output.
 
Such a test against a byte array corresponding to the 
* It can detect Big/Little-endian problems (but only if run on a big endian machine)
* It would fail if encoding is changed in some other way (until expected binary content is updated).

test/Nerdbank.MessagePack.Tests/BuiltInConverterTests.cs
```diff
@@ -101,6 +258,27 @@ public void Guid_MissingHexCharacter([CombinatorialMemberData(nameof(GuidStringF
...
+		Assert.SkipUnless(BitConverter.IsLittleEndian, "This test is written assuming little-endian.");
+		this.Serializer = this.Serializer.WithGuidConverter(OptionalConverters.GuidFormat.LittleEndian);
+
+		Span<Guid> valueSpan = [System.Guid.NewGuid()];
+		Sequence<byte> seq = new();
+		MessagePackWriter writer = new(seq);
+		writer.WriteMapHeader(1);
+		writer.Write(nameof(HasGuid.Value));
+
+		writer.Write(MemoryMarshal.Cast<Guid, byte>(valueSpan));
```
Daniel-Svensson, line 285: Really nice solution which also works for .net472 

I was thinking about maybe using `guid.TryWriteBytes(writer.GetSpan(16));` since it is not limited toLittleEndian, but that is only for NET so it would not have worked for Framework

test/Nerdbank.MessagePack.Tests/BuiltInConverterTests.cs
```diff
@@ -44,10 +126,85 @@ public partial class BuiltInConverterTests : MessagePackSerializerTestBase
...
+		this.AssertRoundtrip(new HasDecimal(new decimal(ulong.MaxValue) * -1000));
+	}
 
+	/// <summary>
+	/// Verifies that we can read <see cref="decimal"/> values that use the Bin header, which is what MessagePack-CSharp's "native" formatter uses.
+	/// </summary>
 	[Fact]
-	public void BigInteger() => this.AssertRoundtrip(new HasBigInteger(1));
+	public void Decimal_FromBin()
+	{
```
Daniel-Svensson, line 149: nit: Maybe you want to do the same as for guid.

```suggestion
	{
		Assert.SkipUnless(BitConverter.IsLittleEndian, "This test is written assuming little-endian.");
```

src/Nerdbank.MessagePack/Converters/PrimitiveConverters.cs
```diff
@@ -295,20 +299,93 @@ internal class DoubleConverter : MessagePackConverter<double>
...
+			? reader.ReadBytes()!.Value
+			: reader.ReadExtension(LibraryReservedMessagePackExtensionTypeCode.ToByte(context.ExtensionTypeCodes.Decimal));
+		if (bytes.Length != sizeof(decimal))
+		{
+			throw new MessagePackSerializationException($"Expected {sizeof(decimal)} bytes but got {bytes.Length}.");
+		}
+
+		Span<byte> decimalBytes = stackalloc byte[sizeof(decimal)];
+		bytes.CopyTo(decimalBytes);
+		decimal result = MemoryMarshal.Read<decimal>(decimalBytes);
```
Daniel-Svensson: nit: In a few other places you have have *IsSingleSegment* optimizations, does it makes sense to do that here as well for consistency ?


```csharp
		decimal result;
		if (bytes.IsSingleSegment)
		{
			result = MemoryMarshal.Read<decimal>(bytes.First.Span);
		}
		else
		{
			Span<byte> decimalBytes = stackalloc byte[sizeof(decimal)];
			bytes.CopyTo(decimalBytes);
			result = MemoryMarshal.Read<decimal>(decimalBytes);
		}
```

src/Nerdbank.MessagePack/Converters/PrimitiveConverters.cs
```diff
@@ -396,155 +444,204 @@ public override void Write(ref MessagePackWriter writer, in decimal value, Seria
...
-				}
-			}
+			throw new MessagePackSerializationException($"Expected {sizeof(ulong) * 2} bytes but got {sequence.Length}.");
 		}
 
-		throw new MessagePackSerializationException("Can't parse to Int128, input string was not in a correct format.");
+		Span<byte> bytesSpan = stackalloc byte[sizeof(ulong) * 2];
+		sequence.CopyTo(bytesSpan);
+		Int128 value = MemoryMarshal.Cast<byte, Int128>(bytesSpan)[0];
+		return BitConverter.IsLittleEndian ? BinaryPrimitives.ReverseEndianness(value) : value;
```
Daniel-Svensson: nit: You could probably have used `BinaryPrimitives.ReadInt128BigEndian` here for slightly cleaner code.
(and `BinaryPrimitives.WriteInt128BigEndian` in write method)

test/Nerdbank.MessagePack.Tests/BuiltInConverterTests.cs
Daniel-Svensson: Or for "HasDecimal" type
```csharp
ReadOnlySpan<byte> byteArray = [ 0x81, 0xA5, 0x56, 0x61, 0x6C, 0x75, 0x65, 0xC4, 0x10, 0x00, 0x00, 0x00, 0x80, 0xE7, 0x03, 0x00, 0x00, 0x18, 0xFC, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF ];
```

Daniel-Svensson (approved):
I've looked at the code, and it looks very good.

I think you nailed all the design decisions and even achieved better inter-op/compatibility (with both binary and string encoding) than I had originally imagined.

I only have a few comments/nits, feel free to address or not address as as you see fit.

Daniel-Svensson:
> Encode Int128 and UInt128 as Bnative integers if they fit, otherwise as a BE msgpack extension. Allow decoding a LE Bin for compatibility with MessagePack-CSharp.

**The current approach of int128 endianness is fine, just wanted to share main train of thoughts and findings when researching**

**Feedback**
I was a bit surprised about, it being BE (mostly since MessagePack-CSharp and BigInteger is LE) but it fits with the MessagePack spec for other primitive types. If the type would ever be added to the standard it is nice to just be able to map it to a negative value and have it work (which seems extremely unlikely since there are a few nice suggestion including ["bigint" support](https://github.com/msgpack/msgpack/pull/249/files)  that would probably been added already).

Note: One interesting aspect of the proposal above, is that it suggest using the same type code for fixed lenght (int 128) and variable length integers (BigInteger). it might save a value but would make it a more complex to handle since it would represent different types depending of type of extension

From my point of view it would be equally valid (I am **not** saying it would be better) to do LE for these reasons:
* to keep it _consistent_ with the other "application" defined types (guid, BigInteger).  
   * (you can treat this as an argument for trying to do BE for BigInteger instead which would need more code and probably different code paths for .NET and netstandard)
* slightly less code (since same code path as MessagePack-CSharp compat)
*  (just speculating) it might be easier to create a joint extension spec for .NET with MessagePack-CSharp 

I did a quick search and could only find a  MessagePack-CSharp  and a[ rust library](https://github.com/3Hren/msgpack-rust/pull/262) which did 128 bit integers natively (and rust did so using BE Bin) so there does not seems to be any common format for other libraries so **_either endianness seems like a valid choice._**


**future thoughts**
 It seems likely we will get [decimal32 decimal64 and decimal128](https://github.com/dotnet/runtime/pull/100729) in .NET 11 .
 If support for them are later added as extensions then we should probably just do EncodeBinary/EncodeDecimal to get corresponding int/long/int128 type and write it with the same endianness as for (U)Int128

test/Nerdbank.MessagePack.Tests/BuiltInConverterTests.cs
AArnott: I like the idea of testing with a value with unique int values. Thanks.

src/Nerdbank.MessagePack/Converters/PrimitiveConverters.cs
AArnott: Sure

AArnott:
Combining BigInteger and Int128 is an interesting idea. All of your ideas and thoughts are good and valid. And yes, I think you identified why I went with BE for the types I did (likelihood to be how msgpack itself specs it someday).

As for using LE for BigInteger and Guid, ... I'm divided. 

There isn't really a precedent in the msgpack spec for whether Guid would be BE or LE. I guess BE, probably. 

BigInteger should probably be BE I suppose. And .NET makes that easy, but .NET Framework does _not_. 

Since there's no one we need to interop with (yet) that wants BE for these types specifically, I'm hedging bets by taking the easy way out and giving the type extension names "LittleEndian" suffixes so that eventually I can fill in with the non-suffixed, BE variety later, on demand.

As to whether BE or LE is enabled 'by default' when BE support is added, therein lies the problem, since that would be a sort of format breaking change if people update Nerdbank.MessagePack and suddenly start emitting a new extension type that an older Nerdbank.MessagePack is expected to be able to read. So I could just keep emitting LE until people set those extension type codes to null and somehow "enable" the extension type codes we suggest for BE encodings of these types... ah... trade-offs.

AArnott:
Ok, I did the work to avoid the forward compatibility concerns: I made `BigInteger` and `Guid` also encode with big endian to match the rest of msgpack integers.

PR #592, opened 2025-09-03
Bump PolyType to 0.60.1

AArnott:
Throw any time we're asked to serialize or compare object graphs that include delegates

### DIFF
--- a/src/Nerdbank.MessagePack/SecureHash/SecureVisitor.cs
+++ b/src/Nerdbank.MessagePack/SecureHash/SecureVisitor.cs
@@ -111,6 +111,10 @@ internal class SecureVisitor(TypeGenerationContext context) : TypeShapeVisitor, 
 	public override object? VisitSurrogate<T, TSurrogate>(ISurrogateTypeShape<T, TSurrogate> surrogateShape, object? state = null)
 		=> new SurrogateSecureEqualityComparer<T, TSurrogate>(surrogateShape.Marshaler, this.GetEqualityComparer(surrogateShape.SurrogateType, state));
 
+	/// <inheritdoc/>
+	public override object? VisitFunction<TFunction, TArgumentState, TResult>(IFunctionTypeShape<TFunction, TArgumentState, TResult> functionShape, object? state = null)
+		=> throw new NotSupportedException("Delegate typed properties cannot be compared.");
+
 	/// <summary>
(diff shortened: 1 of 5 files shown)


src/Nerdbank.MessagePack/SecureHash/SecureVisitor.cs
```diff
@@ -111,6 +111,10 @@ where property.HasGetter
 	public override object? VisitSurrogate<T, TSurrogate>(ISurrogateTypeShape<T, TSurrogate> surrogateShape, object? state = null)
 		=> new SurrogateSecureEqualityComparer<T, TSurrogate>(surrogateShape.Marshaler, this.GetEqualityComparer(surrogateShape.SurrogateType, state));
 
+	/// <inheritdoc/>
+	public override object? VisitFunction<TFunction, TArgumentState, TResult>(IFunctionTypeShape<TFunction, TArgumentState, TResult> functionShape, object? state = null)
+		=> throw new NotSupportedException("Delegate typed properties cannot be compared.");
```
eiriktsarpalis, line 116: Using reference equality comparison wouldn't be entirely unreasonable IMO. If you still believe it should throw, consider hiding the exception behind a throwing equality comparer so that it only gets exercised if a delegate is a encountered in the object graph.

src/Nerdbank.MessagePack/StandardVisitor.cs
```diff
@@ -735,6 +735,10 @@ string PrepareExceptionMessage()
 	public override object? VisitSurrogate<T, TSurrogate>(ISurrogateTypeShape<T, TSurrogate> surrogateShape, object? state = null)
 		=> new SurrogateConverter<T, TSurrogate>(surrogateShape, this.GetConverter(surrogateShape.SurrogateType, state: state));
 
+	/// <inheritdoc/>
+	public override object? VisitFunction<TFunction, TArgumentState, TResult>(IFunctionTypeShape<TFunction, TArgumentState, TResult> functionShape, object? state = null)
+		=> throw new NotSupportedException("Delegate types cannot be serialized.");
```
eiriktsarpalis, line 740: Before this change, delegate types would get serialized as empty objects. This is also how other non-serializable types like `CancellationToken`, `Task`, or `Stream` are being handled today.

PR #597, opened 2025-09-06
Avoid loading unneeded assemblies while looking up primitive converters

AArnott:
This avoids loading `System.Numerics` and `System.Drawing` during serialization of types that wouldn't require them.

Note that due to https://github.com/eiriktsarpalis/PolyType/issues/252, if either of those assemblies are included in *any* source generated type shapes, they'll be loaded anyway.


AArnott:
Go figure. The string-based approach defeats trimming, and the Type-based approach defeats assembly lazy load.

AArnott:
Turns out the winning combination is... a combination.

Start with string checks to avoid assembly loads, then _also_ do a type check so that trimming works.
And all this has to happen in _generic_ methods or trimming doesn't work.

PR #635, opened 2025-09-16
Adds and enhances built-in property name transformations

### LINKED ISSUE #634: Enhance built-in property name transformations
> Consider adapting the tokenization algorithm used in System.Text.Json for snake case and kebab case transformations:
> 
> https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/Common/JsonSeparatorNamingPolicy.cs
> 
> It can also be applied to camel case transformations, but we had already shipped a naive implementation at the time we could no longer change. 

 _Originally posted by @eiriktsarpalis in [#625](https://github.com/AArnott/Nerdbank.MessagePack/issues/625#issuecomment-3298840072)_

AArnott:
This adds snake and kebab case transforms.
It also re-implements camel and pascal case transforms to be token based for better results (which means schema breaking change to those using the old implementation).

Closes #634


AArnott:
@eiriktsarpalis I trivially adapted the STJ tokenizer to support camel and pascal case by making the separator optional. but the results aren't great. I suspect my adaptation was too naive. Any suggestions?

eiriktsarpalis:
What issues did you encounter?

AArnott:
A bunch of unexpected case changes. I think it's basically that your tokenizing method expects a separator between 'words' so that all characters can be upper or lower case. But when there is no separator, we want to switch case for the next character. There's no facility to do that with the original STJ method.

AArnott:
I fixed it (arguably). I'd love to get approval on the camelCase and PascalCase test cases.

test/Nerdbank.MessagePack.Tests/MessagePackNamingPolicyTests.cs
```diff
@@ -0,0 +1,352 @@
...
+	[InlineData("Url", "url")]
+	[InlineData("Id", "id")]
+	[InlineData("I", "i")]
+	[InlineData("Person", "person")]
+	[InlineData("IPhone", "iPhone")]
+	[InlineData("IPhone", "i Phone")]
+	[InlineData("MyUrl", "myURL")]
+	[InlineData("MyUrlValue", "myURLValue")]
+	[InlineData("AlreadyPascal", "AlreadyPascal")]
+	[InlineData("This_is_a_test", "THIS_IS_A_TEST")]
```
AArnott, line 63: This one seems debatable.

PR #654, opened 2025-09-25
Ignore non-serializable types (e.g. delegates) buried in properties that we won't serialize

AArnott:
We use new converter monads to allow representing success and failure cases, so that failure cases can be tracked and thrown or ignored by the visitor callers as appropriate. This allows us to completely avoid throwing exceptions until we know that we want to throw an observed one.

### DIFF
--- a/src/Nerdbank.MessagePack/Result`2.cs
+++ b/src/Nerdbank.MessagePack/Result`2.cs
@@ -1,0 +1,157 @@
+﻿// Copyright (c) Andrew Arnott. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+
+namespace Nerdbank.MessagePack;
+
+/// <summary>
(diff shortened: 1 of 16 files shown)


src/Nerdbank.MessagePack/Result`2.cs
```diff
@@ -0,0 +1,157 @@
...
+namespace Nerdbank.MessagePack;
+
+/// <summary>
+/// Represents the outcome of an operation that may have produced either a value of type <typeparamref name="TValue"/>
+/// or an exception describing a failure.
+/// </summary>
+/// <typeparam name="TValue">The type of the successful result value.</typeparam>
+/// <typeparam name="TError">The type of the error value.</typeparam>
+[DebuggerDisplay($"{{{nameof(DebuggerDisplay)},nq}}")]
+internal struct Result<TValue, TError> : IEquatable<Result<TValue, TError>>
```
eiriktsarpalis, line 15: Someday, we'll be able to express this as a proper union type in C# :-)

PR #665, opened 2025-09-27
Switch to NuGet Trusted Publishing

AArnott:


AArnott:
@micheloliveira-com I wonder if you happen to have any insights into why this doesn't work. I essentially copied it from the other repos of mine that you contributed to, but my first attempt at publishing a package after making this change (across any of my repos) [failed](https://github.com/AArnott/Nerdbank.MessagePack/actions/runs/18117565257/job/51559073075).

micheloliveira-com:
@AArnott Hi,  

It looks like the issue might be with the `NUGET_USER` secret.  

That value needs to match the user who created the policy.  

See details here: [std-uritemplate/std-uritemplate#375](https://github.com/std-uritemplate/std-uritemplate/pull/375)  

https://github.com/std-uritemplate/std-uritemplate/pull/375#issuecomment-3351930128

AArnott:
@micheloliveira-com Thanks. The problem turned out to be that my Trusted Publishing registration on nuget.org had a workflow file path of `.github/workflows/release.yml`, where it should have been simply `release.yml`.

PR #669, opened 2025-09-27
Test chaining use of aggregating type shape provider

AArnott:


AArnott:
This test currently fails, likely due to the same issue discussed in https://github.com/eiriktsarpalis/PolyType/issues/305.
And unblocked by eiriktsarpalis/PolyType#306.

PR #696, opened 2025-10-14
More JIT optimizations

AArnott:


src/Nerdbank.MessagePack/StandardVisitor.cs
```diff
@@ -105,29 +105,37 @@ ConverterResult NonPrimitiveObjectHelper()
 						KeyAttribute? keyAttribute = (KeyAttribute?)property.AttributeProvider?.GetCustomAttributes(typeof(KeyAttribute), false).FirstOrDefault();
 						if (keyAttribute is not null || this.owner.PerfOverSchemaStability || objectShape.IsTupleType)
 						{
-							propertyAccessors ??= new();
-							int index = keyAttribute?.Index ?? propertyIndex;
-							while (propertyAccessors.Count <= index)
+							UsesKeys();
+							void UsesKeys()
```
eiriktsarpalis, line 109: Should this be annotated with `NoInlining` just in case?

AArnott: I tried both ways. It made no difference.

PR #749, opened 2025-11-06
Improve ASP.NET Core MVC doc

AArnott:


samples/AspNetMvc/Controllers/MainController.cs
```diff
@@ -0,0 +1,18 @@
...
+#region Controller
+public class MainController : Controller
+{
+    [Produces("application/x-msgpack")]
+    public IActionResult Index()
+    {
+        this.ViewBag.Message = "Hello World from ASP.NET Core MVC!";
+        return this.View();
+    }
+}
```
amirreza-sf80:
```suggestion
public class MainController : ControllerBase
{
    [Produces("application/x-msgpack")]
    public IActionResult Index()
    {
        return Ok("Hello World from ASP.NET Core MVC!");
    }
}
```

it should probably be an API like this

samples/AspNetMvc/Program.cs
```diff
@@ -18,6 +19,8 @@ void ConfigureServices(IServiceCollection services)
     });
 }
 
-[GenerateShapeFor<bool>] // add an attribute for each top-level type that must be serializable
+// Add an attribute for each top-level type that must be serializable,
+// as well as for ProblemDetails so that error responses can be serialized.
+[GenerateShapeFor<ProblemDetails>]
```
amirreza-sf80, line 36: maybe one other example would be good

AArnott:
@amirreza8002 I applied your suggestions (a little differently than you proposed). Can you take another look?

samples/AspNetMvc/Controllers/MainController.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Andrew Arnott. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.AspNetCore.Mvc;
+using PolyType;
+
+namespace AspNetMvc.Controllers;
+
+#region Controller
+public partial class MainController : ControllerBase
```
amirreza-sf80: doesn't need to be partial

samples/AspNetMvc/Controllers/PersonController.cs
```diff
@@ -0,0 +1,29 @@
...
+    public partial record Person(int Id, string Name);
+
+    // Add an attribute for each top-level type that must be serializable
+    // That does not have have its own [GenerateShape] attribute on it.
+    // Here, we add `int` because it's taken as a parameter type for an action.
+    // Although strictly speaking `int` is already covered implicitly because it
+    // also appears as a property on Person.
+    [GenerateShapeFor<int>]
+    partial class Witness;
+}
```
amirreza-sf80, line 42: codes that I've seen don't put these in the controller class
it's usually out side of it (and in another file, but we don't need that to give examples

samples/AspNetMvc/Controllers/MainController.cs
AArnott: It _does_, actually. Because there are partial nested types (which need to be partial for source generated types).

samples/AspNetMvc/Controllers/PersonController.cs
AArnott: Thanks for the feedback. In other cases like System.Text.Json, compilation breaks or generated code is redundant if you have multiple of these 'witness' types. But with PolyType, you don't have either problem. Keeping them 'local' to where they're used helps users maintain the list so that it doesn't grow in some central place, never to be removed again, forever bloating your application.
So while I accept what you say in that this isn't the common pattern, I propose that it is the preferred pattern going forward for PolyType-based apps.

amirreza-sf80: is Witness needed when we use `GenerateShape` on top of the dto?

and does the dto need to be inside the controller when `GenerateShape` is used on it?

AArnott: We only need the witness type for types that cannot be directly annotated with `[GenerateShape]`.
The DTO can be declared anywhere. For purposes of a simple sample that includes what we want in a small amount of code, I declared them as nested types.

PR #769, opened 2025-11-14
Update to PolyType 1.0.0-rc.8

AArnott:
Also take advantage of the new, faster APIs.

### DIFF
--- a/src/Nerdbank.MessagePack/Converters/ObjectConverterBase`1.cs
+++ b/src/Nerdbank.MessagePack/Converters/ObjectConverterBase`1.cs
@@ -3,9 +3,7 @@
 
 using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
-using System.Reflection;
 using System.Text.Json.Nodes;
-using PolyType.Utilities;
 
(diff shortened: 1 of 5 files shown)


src/Nerdbank.MessagePack/Converters/ObjectConverterBase`1.cs
```diff
@@ -21,7 +19,7 @@ internal abstract class ObjectConverterBase<T> : MessagePackConverter<T>
 	/// <param name="attributeProvider">The attribute provider for the target.</param>
 	/// <param name="schema">The schema for the target.</param>
 	/// <param name="namePrefix">An optional prefix to include in the description, or to use by itself when no <see cref="DescriptionAttribute"/> is present.</param>
-	protected static void ApplyDescription(ICustomAttributeProvider? attributeProvider, JsonObject schema, string? namePrefix = null)
+	protected static void ApplyDescription(IGenericCustomAttributeProvider? attributeProvider, JsonObject schema, string? namePrefix = null)
```
eiriktsarpalis: Note that attribute providers on shapes are now always non-nullable so you might be able to update the nullability here as well.

PR #792, opened 2025-12-01
Add `ValidationProblemDetails` serialization to sample `Witness` class

amirreza-sf80:
there's always a new surprise with asp :)

it caused problems when a validation attribute failed

```cs
    [SocialId]
    public string SocialId { get; set; } = null!;
```

where SocialId is

```cs
[AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
public class SocialIdAttribute() : ValidationAttribute
```


amirreza-sf80:
ah
forgot to turn off formatter
i'll fix it

PR #836, opened 2025-12-26
Call out witness pattern on unions doc

### LINKED ISSUE #835: Provide example of how to (de)serialize with runtime derived type registration
Following [this](https://aarnott.github.io/Nerdbank.MessagePack/docs/unions.html?tabs=net#runtime-derived-type-registration) example, it is not clear to me how I am supposed to (de)serialize instances of the base type. The base type will not be an `IShapeable`, so calling the regular `Serialize(obj)` will result in an error. Adding some clarification in the docs (and/or here) would be appreciated.

AArnott:
Closes #835

### DIFF
--- a/docfx/docs/unions.md
+++ b/docfx/docs/unions.md
@@ -200,6 +200,12 @@ Consider the following example where a type hierarchy is registered without usin
 
 ---
 
+Runtime derived type registration is no substitute for requiring [type shapes](type-shapes.md).
+Rather, runtime derived type registration is useful when:
+
+1. The shapes do not express any derived type information.
+1. You need to override the derived type information in the shapes, due to it being incorrect or incomplete (e.g. your application is extensible and you don't have complete derived type information at compile-time).
+
 ### Duck typing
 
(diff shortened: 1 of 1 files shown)


CptWesley:
@AArnott not exactly. This added comment seems to defeat the entire point of "Runtime derived type registration". I did actually manage to get it to work by calling the `Serialize` and `Deserialize` overloads that take a shape and passing a reflection-generated shape for the base class after setting up the derived type mappings as described in the "Runtime derived type registration" section.

AArnott:
Thanks for the feedback. Runtime derived type registration is no substitute for type shapes. Types must always have shapes. You can get those shapes via attributes (directly or via witnesses) or via the reflection type shape provider.

Those shapes _may_ indicate proper derived type information. If they do, you don't need the 'runtime derived type registration' feature. But if the shapes don't (whether the attributes are missing, or you want to rewrite how it works), then that's what the runtime derived type registration APIs are for.
I'll see if I can improve the docs to indicate this.

Issue #884, opened 2026-02-11
Mixing `[DerivedTypeShape]` and `[TypeShape(Marshaler = )]` drops derived type information


AArnott:
### Discussed in https://github.com/AArnott/Nerdbank.MessagePack/discussions/881

<div type='discussions-op-text'>

<sup>Originally posted by **AlleSchonWeg** February 10, 2026</sup>
Hi,
i have type which inherits from a base class. The base class and the derived class have a Marshaler:
```C#
  [DerivedTypeShape(typeof(DerivedClass), Tag = 1, Name = "DerivedClass")]
  [TypeShape(Marshaler = typeof(BaseClassMarshaler))]
  public class BaseClass
  {
      public string BaseProperty { set; get; }
  }

  [TypeShape(Marshaler = typeof(DerivedClassMarshaler))]
  public class DerivedClass : BaseClass
  {
      public string DerivedProperty { set; get; }
  }

  [GenerateShape]
  public partial class Container
  {
      public BaseClass Base { set; get; }
  }
```
And the IMarshaler implementations:
```C#
public class BaseClassMarshaler : IMarshaler<BaseClass, string>
{
    public string? Marshal(BaseClass? value)
    {
        Console.WriteLine("Marshaling BaseClass to string");
        return "To_MarshalledType";
    }
    public BaseClass? Unmarshal(string? value)
    {
        return new BaseClass();
    }
}

public class DerivedClassMarshaler : IMarshaler<DerivedClass, string>
{
    public string? Marshal(DerivedClass? value)
    {
        Console.WriteLine("Marshaling DerivedClass to string");
        return "To_MarshalledType";
    }
    public DerivedClass? Unmarshal(string? value)
    {
        return new DerivedClass();
    }
}
```
Th problem is on serialization:
```C#
var container = new Container()
{
    Base = new DerivedClass()
    {
        BaseProperty = "BaseValue",
        DerivedProperty = "DerivedValue"
    }
};  
var typeshape = TypeShapeProvider_Nb_MsgPack_Surrogate.Default.GetTypeShape(typeof(Container));
var bytes = serializier.SerializeObject(container, typeshape);
```
the DerivedClassMarshaler is never called. Only the BaseClassMarshaler. Is this a bug? I need  different surrogate types for BaseClass and DerivedClass. </div>

AArnott:
[...]

PR #885, opened 2026-02-11
Document how union and surrogate types interact

AArnott:
* [x] We need to squash this PR, or rebase it to eliminate the version.json changes.
* [x] We should include a sample in the docs demonstrating how to actually mix these two, since it's definitely not trivial.


AArnott:
I have concerns with my approach here. I haven't thoroughly reviewed the (Copilot generated) code yet, since I'm thinking PolyType might be a better place for fixing this. I've shared these concerns with @eiriktsarpalis in #884.

